PR #7, opened 2023-08-03
Add WSL2 note for Win

philipp-baumann:


philipp-baumann:
@b-rodrigues this is just a short note. What we could also add is a link to how to install WSL2?

philipp-baumann:
Then, as you mentioned, probably it is easiest to bootstrap everything using nix on WSL2 ubuntu or whatever.

PR #39, opened 2023-08-12
- `nix_build()`: fix defensive check message

philipp-baumann:
@b-rodrigues I just noticed that the defensive check error in `nix_build()` was still printing a `nix_file` quoting error message. This fixes accordingly the message using `project_path` and `default.nix` in the message.

Since this does not affect the behavior of `nix_build()` but rather just is the error message, I think we do not need to increment versioning using "patch". However, if you want to increment to 0.1.2, I am very happy to do so and update news accordingly


b-rodrigues:
thanks! let's leave it at 0.1.1

PR #45, opened 2023-08-12
README: change because no R shellHook

philipp-baumann:
guess we need to change this since there is no R shellHook in newest `inst/extdata/default.nix`


b-rodrigues:
no actually we want the shellhook

Issue #56, opened 2023-08-16
`inst/extdata/default.nix` use `--vanilla` flag for R startup 


philipp-baumann:
related to #54 
We currently get an error when running `nix-shell ./inst/extdata/default.nix` in `nix.yaml` CI. I believe it is more consistent if we use a plain R session that does not read user configuration files such as `.Rprofile`, because this could pollute stuff later and it becomes harder to diagnose issues of users.

PR #57, opened 2023-08-16
`default.nix` use `--vanilla` to fix nix R CI shell error

philipp-baumann:
addresses #56 ; adapt shellHook


philipp-baumann:
- use `R --vanilla` in shellHook to not propagate user-specific `.Renviron` and `.Rprofile`. Addresses #56

PR #83, opened 2023-09-26
fix locale warnings by setting envvars in nix

### LINKED ISSUE #50: Find a way to fix R `locale` in Nix for `darwin` and `linux`
This is to have a universal clean solution, since we did a quick-fix for #48

philipp-baumann:
fixes #50


b-rodrigues:
I guess in some sense setting the locale explicitly to en_US is better for reproducibility, and making it appear explicitly in the default.nix could nudge users to set it explicitly to whichever locale they prefer. Let’s keep this for now as it seems to solve more issues than alternatives.

PR #109, opened 2024-01-26
Documentation feedback

philipp-baumann:


README.Rmd
```diff
@@ -29,35 +29,43 @@ knitr::opts_chunk$set(
...
+version of R and R packages (as well as other tools or languages, if needed).
+You can use `{rix}` and Nix to replace `{renv}` and Docker with one single tool.
+Nix is an incredibly useful piece of software for ensuring reproducibility of
+projects, in research or otherwise. For example, it allows you run web 
+applications like Shiny apps or `{plumber}` APIs in a controlled environment.
+
+Nix has a fairly high entry cost though. Nix is a complex piece of software that
+comes with its own programming language, which is also called Nix. Its purpose
+is to solve a complex problem: defining instructions how to build software
+packages and manage configurations in declarative way. This makes sure that
```
b-rodrigues, line 44: a declarative way

README.Rmd
```diff
@@ -29,35 +29,43 @@ knitr::opts_chunk$set(
...
+possible to create project-specific environments that contain a project-specific
+version of R and R packages (as well as other tools or languages, if needed).
+You can use `{rix}` and Nix to replace `{renv}` and Docker with one single tool.
+Nix is an incredibly useful piece of software for ensuring reproducibility of
+projects, in research or otherwise. For example, it allows you run web 
+applications like Shiny apps or `{plumber}` APIs in a controlled environment.
+
+Nix has a fairly high entry cost though. Nix is a complex piece of software that
+comes with its own programming language, which is also called Nix. Its purpose
+is to solve a complex problem: defining instructions how to build software
```
b-rodrigues, line 43: on how to

README.Rmd
```diff
@@ -29,35 +29,43 @@ knitr::opts_chunk$set(
...
+hardware.
+
+`{rix}` provides functions to help you write and deploy Nix expressions (written
+in the Nix language). These expressions will be the inputs for the Nix package
+manager, to build sets of software packages and provide them in a reproducible
+development environment. These environments can be used for interactive data
+analysis, or reproduced when running pipelines in CI/CD systems. On the 
+[Nixpkgs collection](https://github.com/nixos/nixpkgs), there are currently more
+than 80.000 pieces of software available through the Nix package manager. Using
+{rix}, you can define and build isolated R environments through Nix package
```
b-rodrigues, line 55: through the Nix

README.Rmd
```diff
@@ -29,35 +29,43 @@ knitr::opts_chunk$set(
...
+manager with ease. Like this, environments contain R and all the required 
+packages that you need for your project. You can also add any other software
+tool available. The Nix R ecosystem currently includes the entirety of CRAN and
+Bioconductor packages. Like with any other programming language and software, it
+is also possible to install older releases of R packages, or install packages
+from GitHub at defined states.
+
+The Nix package manager is extremely powerful. Not only does it handle all the
+dependencies their builds extremely well in deterministic manner. It is also
```
b-rodrigues, line 64: Not only does it handle all the dependencies of any package extremely well in a deterministic manner, it is also possible with it to reproduce environments containing old releases of software.

I would write it like this, just one sentences. But here we repeat that it handles old software releases. Maybe we should remove it above?

README.Rmd
```diff
@@ -68,9 +76,10 @@ WSL2) and `{rix}` comes with the following features:
 - have several versions of R and R packages installed at the same time on the
   same system;
 - define complete development environments as code and use them anywhere;
-- run single function in a different environment (potentially with a different R
-  version and R packages) for an interactive R session and get back the output
-  of that function using `with_nix()`;
+- run single R function (and objects in the call stack) in a different
```
b-rodrigues, line 79: single R functions (plural)

README.Rmd
```diff
@@ -151,11 +163,14 @@ nix-shell --expr "$(curl -sl https://raw.githubusercontent.com/b-rodrigues/rix/m
 
 You can then create new development environment definitions, build them, and
 start using them.
+</details>
 
 ## Getting started for new users
 
 To get started with `{rix}` and Nix, you should read the following vignette
-`vignette("a-getting-started")`. The vignettes are numbered to get you to learn
+`vignette("a-getting-started")`
```
b-rodrigues, line 171: we’re repeating the link here, no? the call to `vignette("a-getting-started")` will be transformed to a link by pkgdown automatically

b-rodrigues (changes requested):

README.Rmd
philipp-baumann: does this work for the README, too? Currently i don't see the link there.

README.Rmd
philipp-baumann: fair point, i'll remove the repetition.

philipp-baumann: back to one sentence sounds good.

README.Rmd
b-rodrigues: ah not on github yes, but on the pkdown’ed website it works. I guess we should add the link back then

philipp-baumann:
@b-rodrigues i first thought we could include something like this:


```sh
"### Path of Nix installation not in `PATH`

When using RStudio Desktop on macOS, you typically launch it from the Applications folder. However, RStudio currently lacks an option to start an integrated R session via a shell startup on macOS (in contrast to Linux, where this is the case). As a result, key environmental variables for UNIX systems, like PATH, are not be properly loaded from your default shell (e.g., .zsh via ~/.zshrc). Also, RStudio overwrites a `PATH` variable set via `.Rprofile` or `.Renviron` with its own defaults, which makes it impossible to find Nix and standard Nix tools like `nix-build`.  It's worth noting that this doesn't impact rix::rix(), which generates R expressions and doesn't require a Nix installation. As a workaround, we have added mechanisms in `nix_build()`, and `with_nix()` that append the path of the Nix store to the `PATH` variable in an active session automatically (via `Sys.setenv()`, when you use RStudio on macOS
```

but then, it might be also quite a bit of an overhead of information, since it gets taken care of without user intervention. Shall we mention still or we leave it, what's your opinion?

b-rodrigues:
Maybe we could add this all the way at the end, with a title like "why do we need this?" or something. And start with a first sentence such as: "only read this if you're interested in the why". This could be interesting info for some users, and also for us in 3 years when we'll have forgotten why we do that :D

philipp-baumann:
addressed in #130

PR #113, opened 2024-01-27
`rix::rix()`: `R --vanilla` -> `R` for shellHook to enable `.Rprofile` inheritance from `rix_init()`

philipp-baumann:


b-rodrigues:
I wonder if we should have a default shell_hook at all

philipp-baumann:
> I wonder if we should have a default shell_hook at all

true :-) maybe people call `Rscript` or any other stuff directly if they already use R. In `with_nix()` we also launch via Rscript and that does not need a shell hook

philipp-baumann:
Will do a separate PR for this and also with the tests explicitly telling.

PR #116, opened 2024-01-28
nix test CI: testthat and covr tests for `with_nix()`

philipp-baumann:


philipp-baumann:
we still have #117 for macOS users; it is something specific to that version of nixpkgs and the aarch64 architecture on darwin. Wonder if we get that for linux and the same architecture, too.

philipp-baumann:
Points addressed:

## `with_nix()`
- [ ] If Nix R evaluation of `expr` fails, then no output should be produced (delete `_out.Rds` to not load outputs from previous runs in the same tempdir into the R session) => Bug
 - [ ] wrap session info in `try()` for Nixpkgs R 3.5.0, because `sessionInfo()` is not working properly

PR #127, opened 2024-02-02
Patch `rix()` with `shell_hook = NULL` default, Nix-R wrappers `nix_build()` and `with_nix()`; update testfiles

philipp-baumann:


philipp-baumann:
Currently, `Rscript devtools::test()` runs tests for `with_nix()` and `nix_build()` successfully, however `covr::package_coverage()` still fails with exit code. We can reproduce on Ubuntu 22.04 VM, however it is not failing in Debian 12.
https://github.com/ropensci-review-tools/pkgcheck/issues/195

PR #136, opened 2024-02-06
added skip on covr on top

b-rodrigues:


b-rodrigues:
`covr::packages_coverage()` passes on my local opensuse install though, but not on GA, even though the problematic tests should get skipped.

Issue #155, opened 2024-04-14
Docs: install rix with with default.nix, not install.packages()


artur-sannikov:
Hi,

I was following [this guide ](https://b-rodrigues.github.io/rix/articles/b1-setting-up-and-using-rix-on-linux-and-windows.html)to install `rix`. I noticed that after we download [default.nix](https://raw.githubusercontent.com/b-rodrigues/rix/master/inst/extdata/default.nix) and then run `rix()`  command, `default.nix` is overwritten (`overwrite = TRUE`), which obviously removes rix after `nix-build`.

I think, `rix` should be always specified in `default.nix` as it's the goal of the project to use `nix-build` for completely reproducible environments. Using 

```R
install.packages("rix", repos = [c](https://rdrr.io/r/base/c.html)("https://b-rodrigues.r-universe.dev",
  "https://cloud.r-project.org"))
```

from the R session partially defeats this purpose.

A better `rix()` command might look like this:

```R
rix(r_ver = "latest",
    r_pkgs = c("dplyr", "ggplot2"),
    git_pkgs = list(package_name = "rix",
                   repo_url = "https://github.com/b-rodrigues/rix",
                   branch_name = "master",
                   commit = "76d1bdd03d78589d399b4b9d473ecde616920a82"),
    ide = "other",
    project_path = ".",
    overwrite = TRUE)
```

b-rodrigues:
Thanks! This was indeed confusing. I believe it should be better now. Could you take a look at PR https://github.com/b-rodrigues/rix/pull/157 ?

PR #157, opened 2024-04-18
Update b1-linux_win.Rmd

b-rodrigues:
Dealing with #155


artur-sannikov:
Hi, yes, it's more clear now. However, 

> If you need to add packages for your project, rerun the command above, but add the needed packages to `r_pkgs`.

I'd add a note: if you run the previous `rix()` command, you will overwrite your `default.nix` file and the next time you access the shell, you won't have `rix` installed. To make sure you have it always installed, include `rix` in the `rix` command itself like shown below.

PR #161, opened 2024-04-29
Add more IDE-s: rserver and radian

Kupac:
Adds support to RStudio server and radian. It is done in a way that avoids core repetition, and allows easy addition of future wrappers.

In addition, {fusen} also modified some files (vignetters, DESCRIPTION), probably for the better, so I also included those in the commit.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -21,6 +21,7 @@ Imports:
     sys,
     utils
 Suggests: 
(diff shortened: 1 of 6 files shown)


b-rodrigues:
Amazing, this looks great!

b-rodrigues:
Many thanks!

PR #163, opened 2024-05-19
Bleeding edge

b-rodrigues:
Added the possibility to use "bleeding_edge" and "frozen_edge" to R versions. This uses the rstats-on-nix fork that provides daily updates to R packages.


b-rodrigues:
Still need to finish writing vignettes, but functionality is done and tests are passing.

b-rodrigues:
Need to think of a good test for the case when `r_version = "frozen_edge"` for rix(), as this changes everytime there's a new commit on rstats-on-nix daily branch.

Issue #184, opened 2024-06-13
Add possibility to install local tarballs of packages


b-rodrigues:
This works, need to make it automated:

```
let
 pkgs = import (fetchTarball "https://github.com/NixOS/nixpkgs/archive/8ad5e8132c5dcf977e308e7bf5517cc6cc0bf7d8.tar.gz") {};
 local_pkgs = [(pkgs.rPackages.buildRPackage {
    name = "chronicler";
    src = ./chronicler_0.2.2.tar.gz;
    propagatedBuildInputs = builtins.attrValues {
      inherit (pkgs.rPackages) 	clipr diffobj dplyr maybe rlang stringr tibble;
    };
  }) ];
  system_packages = builtins.attrValues {
  inherit (pkgs) R glibcLocales nix pandoc;
};
  in
  pkgs.mkShell {
    LOCALE_ARCHIVE = if pkgs.system == "x86_64-linux" then  "${pkgs.glibcLocales}/lib/locale/locale-archive" else "";
    LANG = "en_US.UTF-8";
    LC_ALL = "en_US.UTF-8";
    LC_TIME = "en_US.UTF-8";
    LC_MONETARY = "en_US.UTF-8";
    LC_PAPER = "en_US.UTF-8";
    LC_MEASUREMENT = "en_US.UTF-8";

    buildInputs = [ local_pkgs system_packages ];
  }
```

PR #191, opened 2024-06-17
Install local

b-rodrigues:
Added possibility to install source packages from local archives. 
Introduced a new dependency to `{desc}` to get the dependencies of the local archive. 

Seems to be working, but requires further/more clever testing. For now, I've added the source to chronicler in the test and `skip_on_cran()` but there might be a better way. Damn these tests that require interaction with the outside world!


b-rodrigues:
Addresses #184

philipp-baumann:
> Added possibility to install source packages from local archives. Introduced a new dependency to `{desc}` to get the dependencies of the local archive.
> 
> Seems to be working, but requires further/more clever testing. For now, I've added the source to chronicler in the test and `skip_on_cran()` but there might be a better way. Damn these tests that require interaction with the outside world!

this is a really handy feature, thanks for implementing! :-) We don't need {desc}, i'll comment directly in the code.

b-rodrigues:
ok so I think I'm going to change strategies: I'm not going to do a unit test for this, instead, I'll write an action that will download two packages, and then generated a Nix expression and then try to build it. If it's all green, then we know we're good

PR #199, opened 2024-06-23
Chores before releasing rix 0.8.0

philipp-baumann:


philipp-baumann:
- [ ] added opinionated lintr config
- [ ] linted R source code files via `styler` Rstudio plugin
- [ ] removed base R pipe in source code
- [ ] wrapping of function docs

Issue #200, opened 2024-06-24
`Error in available_r() : object 'r_nix_revs' not found` if the package is not loaded


eliocamp:
`rix::rix:()` seems to fail if the package is not loaded:

``` r
path_default_nix <- tempdir()

rix::rix(r_ver = "latest",
    r_pkgs = c("dplyr", "ggplot2"),
    system_pkgs = NULL,
    git_pkgs = NULL,
    ide = "code",
    project_path = path_default_nix,
    overwrite = TRUE,
    print = TRUE)
#> Error in available_r(): object 'r_nix_revs' not found

library(rix)
rix::rix(r_ver = "latest",
         r_pkgs = c("dplyr", "ggplot2"),
         system_pkgs = NULL,
         git_pkgs = NULL,
         ide = "code",
         project_path = path_default_nix,
         overwrite = TRUE,
         print = TRUE)
#> # This file was generated by the {rix} R package v0.7.1 on 2024-06-24
#> # with following call:
#> # >rix::rix(r_ver = "9693852a2070b398ee123a329e68f0dab5526681",
#> #  > r_pkgs = c("dplyr",
#> #  > "ggplot2"),
#> #  > system_pkgs = NULL,
#> #  > git_pkgs = NULL,
#> #  > ide = "code",
#> #  > project_path = path_default_nix,
#> #  > overwrite = TRUE,
#> #  > print = TRUE)
#> # It uses nixpkgs' revision 9693852a2070b398ee123a329e68f0dab5526681 for reproducibility purposes
#> # which will install R version latest.
#> # Report any issues to https://github.com/b-rodrigues/rix
#> let
#>  pkgs = import (fetchTarball "https://github.com/NixOS/nixpkgs/archive/9693852a2070b398ee123a329e68f0dab5526681.tar.gz") {};
#>  
#>   rpkgs = builtins.attrValues {
#>     inherit (pkgs.rPackages) 
#>       dplyr
#>       ggplot2
#>       languageserver;
#>   };
#>     
#>   system_packages = builtins.attrValues {
#>     inherit (pkgs) 
#>       R
#>       glibcLocales
#>       nix;
#>   };
#>   
#> in
#> 
#> pkgs.mkShell {
#>   LOCALE_ARCHIVE = if pkgs.system == "x86_64-linux" then  "${pkgs.glibcLocales}/lib/locale/locale-archive" else "";
#>   LANG = "en_US.UTF-8";
[...]

PR #217, opened 2024-06-24
Available R

b-rodrigues:
works


philipp-baumann:
That works :-) , @b-rodrigues strange now we also have 403 error on vignette build in Ubuntu devel and not just macOS. really strange; lets reach out to r-lib/actions

b-rodrigues:
fixes #200

PR #220, opened 2024-06-24
use internal lazy-loading; `sysdata.rda` is supposed to be in `R/` and not in `data/`

philipp-baumann:


philipp-baumann:
@b-rodrigues there is now quite a mix; i suggest we either call it `sydata.rds` and put it in `R/sysdata.rds`, which is recommended from CRAN (no qualified call) -> this PR. Or we go via a data approach. the latter i would not do because we want people to get it via `rix::available_r()` and not via data(...)

b-rodrigues:
thanks, that’s much better!

PR #228, opened 2024-06-28
Handling situation where RStudio is chosen but no R packages are added

b-rodrigues:
I think that we should not allow this at all, because the rstudioWrapper has no packages to wrap, and as such, doesn't get installed. I also doubt that anyone would really want RStudio but no packages. If you really only need the R interpreter, you like will not want RStudio either?


philipp-baumann:
> I think that we should not allow this at all, because the rstudioWrapper has no packages to wrap, and as such, doesn't get installed. I also doubt that anyone would really want RStudio but no packages. If you really only need the R interpreter, you like will not want RStudio either?

I agree it is a special and rather rare case. Like suggested https://github.com/b-rodrigues/rix/issues/214, I would put `rstudio` entry to `system_packages` in `default.nix`. I don't see a reason that prevents us from not allowing something special with small effort. It is rare but if one just wants base R and we want to test basic functionality of rstudio (which does a lot of stuff like modifying variables).

b-rodrigues:
I think it should error and tell the user to add rstudio to system_pkgs manually instead. I think it would make it clear that what the user is attempting is not recommended

philipp-baumann (approved):

PR #231, opened 2024-06-28
Run rix_init after rix

b-rodrigues:
This PR would run rix_init each time rix is executed


b-rodrigues:
we could also add this: https://github.com/b-rodrigues/rix/issues/230#issuecomment-2196879617

b-rodrigues:
ready for review

philipp-baumann:
> ready for review
> 
>     * rix_init() is now always called when using rix() with create_missing
> 
>     * rix() creates the project path folder if it's missing
> 
>     * install.packages() is redefined in the .Rprofile if running inside a Nix session to throw an error
> 
>     * documentation was adapted

@b-rodrigues i have to think about it and sleep over it. We really should recommend users to make use of `rix::rix_init()`, but i see a couple of problems mixing it. Will respond a bit later with concrete examples.

philipp-baumann:
todos like discussed on matrix :-)

- [ ] add `message_type` option and also forward it to `rix_init()` called.

b-rodrigues:
> todos like discussed on matrix :-)
> 
> * [x]  add `message_type` option and also forward it to `rix_init()` called.

Implemented, but with a twist: "verbose" is too chatty when rix_init() is called by rix(), so if the user chooses "verbose", it gets "downgraded" to "simple".

philipp-baumann:
> > todos like discussed on matrix :-)
> > 
> > * [x]   add `message_type` option and also forward it to `rix_init()` called.
> 
> Implemented, but with a twist: "verbose" is too chatty when rix_init() is called by rix(), so if the user chooses "verbose", it gets "downgraded" to "simple".

That doesn't make sense to me. If I choose"verbose", I explicitly want "verbose", and consistent behavior across all functions, here `rix_init()`. The default is `"simple"`, and one either wants to have all outputs or none as an alternative.

b-rodrigues:
> That doesn't make sense to me. If I choose"verbose", I explicitly want "verbose", and consistent behavior across all functions, here `rix_init()`. The default is `"simple"`, and one either wants to have all outputs or none as an alternative.

The issue is that "verbose" for rix_init() prints many lines that push the lines from rix() out of the console. So my thinking was that if I'm using "verbose" with rix(), I'm more interested to what rix() has to say. I feel like then having rix_init() message_type set to "simple" would be enough for "verbose" for rix().

But if I'm calling rix_init() directly with "verbose", then yeah, I'd like to know everything rix_init() has to tell me.

philipp-baumann:
@b-rodrigues currently, we still don't get the expected behavior if we already have an `.Rprofile` that was generated before outside of nix. How would we know if a user called `rix_init()` before or not. One might have just a custom `.Rprofile` but without the necessary code to limit `.libPaths()` to nix store paths. In one situation it would work while in the second it gets overlooked and we still have interference from the user library.

b-rodrigues:
> @b-rodrigues currently, we still don't get the expected behavior if we already have an `.Rprofile` that was generated before outside of nix. How would we know if a user called `rix_init()` before or not. One might have just a custom `.Rprofile` but without the necessary code to limit `.libPaths()` to nix store paths. In one situation it would work while in the second it gets overlooked and we still have interference from the user library.

Right, we need to check for existence of an already existing .Rprofile!

- [ ] Check for existence of already existing .Rprofile and warn user.

philipp-baumann:
> > @b-rodrigues currently, we still don't get the expected behavior if we already have an `.Rprofile` that was generated before outside of nix. How would we know if a user called `rix_init()` before or not. One might have just a custom `.Rprofile` but without the necessary code to limit `.libPaths()` to nix store paths. In one situation it would work while in the second it gets overlooked and we still have interference from the user library.
> 
> Right, we need to check for existence of an already existing .Rprofile!
> 
>     * [ ]  Check for existence of already existing .Rprofile and warn user.

how do you know its generated?

> > That doesn't make sense to me. If I choose"verbose", I explicitly want "verbose", and consistent behavior across all functions, here `rix_init()`. The default is `"simple"`, and one either wants to have all outputs or none as an alternative.
> 
> The issue is that "verbose" for rix_init() prints many lines that push the lines from rix() out of the console. So my thinking was that if I'm using "verbose" with rix(), I'm more interested to what rix() has to say. I feel like then having rix_init() message_type set to "simple" would be enough for "verbose" for rix().
> 
> But if I'm calling rix_init() directly with "verbose", then yeah, I'd like to know everything rix_init() has to tell me.

rix::rix() does two things and I would want to exactly know what it does. I see two options, either we delete 

> > @b-rodrigues currently, we still don't get the expected behavior if we already have an `.Rprofile` that was generated before outside of nix. How would we know if a user called `rix_init()` before or not. One might have just a custom `.Rprofile` but without the necessary code to limit `.libPaths()` to nix store paths. In one situation it would work while in the second it gets overlooked and we still have interference from the user library.
> 
> Right, we need to check for existence of an already existing .Rprofile!
> 
>     * [ ]  Check for existence of already existing .Rprofile and warn user.

Why don't we add a `rprofile_action` argument to `rix::rix()`. We default to `create_missing`, but also signal other options. How do we know whether `rix::rix_init()` `.Rprofile` is taking action?

philipp-baumann:
- [ ]  for `message_type = "quiet"`, it should not warn and message:

```r
> rix(
+     r_ver = "latest",
+     r_pkgs = c("dplyr", "ggplot2"),
+     system_pkgs = NULL,
+     git_pkgs = NULL,
+     local_pkgs = NULL,
+     ide = "code",
+     project_path = "_test",
+     overwrite = TRUE,
+     print = FALSE,
+     message_type = "quiet",
+     shell_hook = NULL
+ )
Warning message:
In rix(r_ver = "latest", r_pkgs = c("dplyr", "ggplot2"), system_pkgs = NULL,  :
  

### .Rprofile file already exists. You may want to call rix_init(profile_action = 'append') manually to ensure correct functioning of your Nix environment. ###
```

philipp-baumann:
- [ ]  profile_action -> rprofile_action. something along "Please see ?rix_init for further details.

b-rodrigues:
> * [ ]   for `message_type = "quiet"`, it should not warn and message:
> 
> ```r
> > rix(
> +     r_ver = "latest",
> +     r_pkgs = c("dplyr", "ggplot2"),
> +     system_pkgs = NULL,
> +     git_pkgs = NULL,
> +     local_pkgs = NULL,
> +     ide = "code",
> +     project_path = "_test",
> +     overwrite = TRUE,
> +     print = FALSE,
> +     message_type = "quiet",
> +     shell_hook = NULL
> + )
> Warning message:
> In rix(r_ver = "latest", r_pkgs = c("dplyr", "ggplot2"), system_pkgs = NULL,  :
>   
> 
> ### .Rprofile file already exists. You may want to call rix_init(profile_action = 'append') manually to ensure correct functioning of your Nix environment. ###
> ```

should we keep this warning, even if quiet ? That message only appears if .Rprofile already exists and wasn't created by rix_init, so it would appear very rarely and I'd say in these cases the users would like to be informed. What do you think ?

philipp-baumann:
> > * [ ]    for `message_type = "quiet"`, it should not warn and message:
> > 
> > ```r
> > > rix(
> > +     r_ver = "latest",
> > +     r_pkgs = c("dplyr", "ggplot2"),
> > +     system_pkgs = NULL,
> > +     git_pkgs = NULL,
> > +     local_pkgs = NULL,
> > +     ide = "code",
> > +     project_path = "_test",
> > +     overwrite = TRUE,
> > +     print = FALSE,
> > +     message_type = "quiet",
> > +     shell_hook = NULL
> > + )
> > Warning message:
> > In rix(r_ver = "latest", r_pkgs = c("dplyr", "ggplot2"), system_pkgs = NULL,  :
> >   
> > 
> > ### .Rprofile file already exists. You may want to call rix_init(profile_action = 'append') manually to ensure correct functioning of your Nix environment. ###
> > ```
> 
> should we keep this warning, even if quiet ? That message only appears if .Rprofile already exists and wasn't created by rix_init, so it would appear very rarely and I'd say in these cases the users would like to be informed. What do you think ?

yeah, you have a good point. Let's keep it like this. Technically it's a warning and it is good to inform at least, while `message_type` infers that it is messaging with `message()` or `cat()` (its fair to use `cat()` like we do in a lot of other places in {rix}. 

Something that just came to my mind: If there is no problem, I would expect that `rix::rix()` gives me a short message that the nix expression was generated and written to`default.nix` at `project_path = <project_path>`.

It would nice to have something like this implemented before we merge; currently, there is none:

```r
> devtools::load_all(".")
ℹ Loading rix
> rix(
+     r_ver = "latest",
+     r_pkgs = c("dplyr", "ggplot2"),
+     system_pkgs = NULL,
+     git_pkgs = NULL,
+     local_pkgs = NULL,
+     ide = "code",
+     project_path = "_test",
+     overwrite = TRUE,
+     print = FALSE,
+     message_type = "simple",
+     shell_hook = NULL
+ )
```

R/rix.R
```diff
@@ -260,9 +292,32 @@ for more details."
...
+    if(!file.exists(.Rprofile_path)){
+      rix_init(project_path = project_path,
+               rprofile_action = "create_missing",
+               # 'verbose' is too chatty for rix()
+               # hence why it's transformed to "simple"
+               message_type = ifelse(message_type == "verbose",
+                                     "simple", message_type))
+
+      if(message_type != "quiet"){
+        message("\n\n### Successfully generated `default.nix` and `.Rprofile` ###\n\n")
```
b-rodrigues, line 307: Regarding your comment bout a successful run, wouldn't this be it ?

philipp-baumann:
after 

```r
> rix(
+     r_ver = "latest",
+     r_pkgs = NULL,
+     system_pkgs = NULL,
+     git_pkgs = NULL,
+     local_pkgs = NULL,
+     tex_pkgs = NULL,
+     ide = c("other", "code", "radian", "rstudio", "rserver"),
+     project_path = ".",
+     overwrite = TRUE,
+     print = FALSE,
+     message_type = "simple",
+     shell_hook = NULL
+ )


### Successfully generated `default.nix`. Keeping `.Rprofile` generated by `rix::rix_init()`###
```

PR #237, opened 2024-07-05
Adding to readme: why Nix and not *insert any other tool here*

b-rodrigues:


philipp-baumann:
great idea Bruno :-) one addition, we should add r2u, {rspm} and {bspm}. e.g. https://cran.r-project.org/web/packages/rspm/readme/README.html https://eddelbuettel.github.io/r2u/ an interesting feature is e.g. `rspm::renv_init()` which bridges renv and automatic system package dependency resolution via apt or dnf/yum.

PR #238, opened 2024-07-05
Removing uneeding branchName argument for github packages

b-rodrigues:
After our discussions and some thought, I think that indeed, we should simply get rid of this argument, and better do it now before submitting to CRAN. I have also changed the code on the server, but not deployed yet; locally, it works!


b-rodrigues:
One we merge this, I'll also merge on the server.

b-rodrigues:
If you want to test the api locally: https://github.com/b-rodrigues/git2nixsha/tree/main

b-rodrigues:
Tests pass locally, but complain here because when nix is not available, they try to reach the api, however the new version without branchName is not deployed it.

philipp-baumann:
> After our discussions and some thought, I think that indeed, we should simply get rid of this argument, and better do it now before submitting to CRAN. I have also changed the code on the server, but not deployed yet; locally, it works!

I agree its a good idea to do this now!

philipp-baumann:
> If you want to test the api locally: https://github.com/b-rodrigues/git2nixsha/tree/main

Can we replace this:

https://github.com/b-rodrigues/git2nixsha/blob/6ad61c8be5f343c6f8851f9617d218a7c013d8ee/git2nixsha.R#L15

by the new approach like we do locally in `hash_url()` and `hash_git`:
https://github.com/b-rodrigues/rix/blob/699372eaadef9a24b7f23121641ce97754fc6659/R/nix_hash.R#L135
https://github.com/b-rodrigues/rix/blob/699372eaadef9a24b7f23121641ce97754fc6659/R/nix_hash.R#L22

Then we really have consistent behavior. To be sure, I think we should also add a new test (CI with nix installed), and compare equality of `get_sri_hash_deps()` test output after `options(rix.hash_nix="api_server")` and tets whether we get identical results compared to calling it after `options(rix.hash_nix="locally"`) (only run if nix is installed).

b-rodrigues:
done with https://github.com/rstats-on-nix/nixpkgs/commit/3fdde11f18b73cc579c841e21cea8f6a8513c65f

Issue #243, opened 2024-07-08
Executing 'nix-hash' failed with status 1


arraytools:
Thanks for providing such a useful software. I have an issue when I try to follow the example on the vignette 'd1 - Installing R packages in a Nix environment' to install an old R package archived on CRAN. Any idea will be appreciated. Thanks.

R version 4.4.1 (2024-06-14) -- "Race for Your Life"
Copyright (C) 2024 The R Foundation for Statistical Computing
Platform: x86_64-pc-linux-gnu

R is free software and comes with ABSOLUTELY NO WARRANTY.
You are welcome to redistribute it under certain conditions.
Type 'license()' or 'licence()' for distribution details.

  Natural language support but running in an English locale

R is a collaborative project with many contributors.
Type 'contributors()' for more information and
'citation()' on how to cite R or R packages in publications.

Type 'demo()' for some demos, 'help()' for on-line help, or
'help.start()' for an HTML browser interface to help.
Type 'q()' to quit R.

> library(rix)
> path_default_nix <- tempdir()

rix(r_ver = "4.2.1",
    r_pkgs = c("dplyr@0.8.0", "janitor@1.0.0"),
    system_pkgs = NULL,
    git_pkgs = NULL,
    ide = "other",
    project_path = path_default_nix,
    overwrite = TRUE)
Error in sys::exec_internal(cmd = cmd, args = args) : 
  Executing 'nix-hash' failed with status 1

> traceback()
17: stop(sprintf("Executing '%s' failed with status %d", cmd, status))
16: sys::exec_internal(cmd = cmd, args = args)
15: nix_sri_hash(path = path_to_source_root)
14: hash_url(url = repo_url)
13: hash_cran(repo_url = repo_url)
12: nix_hash(repo_url, commit)
11: get_sri_hash_deps(repo_url, branch_name = NULL, commit = NULL)
10: FUN(X[[i]], ...)
9: lapply(archive_pkgs, fetchzip)
8: paste(lapply(archive_pkgs, fetchzip), collapse = "\n")
7: fetchzips(archive_pkgs)
6: paste(fetchgits(git_pkgs), fetchzips(archive_pkgs), collapse = "\n")
5: fetchpkgs(git_pkgs, archive_pkgs)
4: sprintf("\n  git_archive_pkgs = [%s  ];\n", fetchpkgs(git_pkgs, 
       archive_pkgs))
[...]

PR #244, opened 2024-07-10
Fix `nix-hash` system call on Debian and Debian-based systems

philipp-baumann:
WIP/needs testing, more details #243


b-rodrigues (approved):
LGTM

PR #285, opened 2024-08-30
Reduce confusion installing nix on Linux

jrosell:
This change try to make a clear recommendation to avoid system package manager and try to avoid people having isues like this  https://discourse.nixos.org/t/nix-var-nix-daemon-socket-socket-permission-denied/39142

### DIFF
--- a/vignettes/b1-setting-up-and-using-rix-on-linux-and-windows.Rmd
+++ b/vignettes/b1-setting-up-and-using-rix-on-linux-and-windows.Rmd
@@ -86,14 +86,18 @@ and then relaunch WSL (Ubuntu) from the start menu.
 Afterwards, you can install Nix like business as usual. You can proceed with the
 Determinate Systems installer.
 
-### The Determinate Systems installer
+### Use the Determinate Systems installeron Linux and Windows
 
 You can use `{rix}` to generate Nix expressions even if you don't have Nix
 installed on your system, but obviously, you need to install Nix if you actually
-want to build the defined development environment and use them. Installing (and
+want to build the defined development environment and use them. 
+
+Installing (and
 uninstalling) Nix is quite simple, thanks to the installer from [Determinate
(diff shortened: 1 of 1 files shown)


vignettes/b1-setting-up-and-using-rix-on-linux-and-windows.Rmd
```diff
@@ -86,14 +86,18 @@ and then relaunch WSL (Ubuntu) from the start menu.
 Afterwards, you can install Nix like business as usual. You can proceed with the
 Determinate Systems installer.
 
-### The Determinate Systems installer
+### Use the Determinate Systems installeron Linux and Windows
```
b-rodrigues, line 89: keep the title as it was before, as the determinate systems installer also works on macOS

vignettes/b1-setting-up-and-using-rix-on-linux-and-windows.Rmd
```diff
@@ -86,14 +86,18 @@ and then relaunch WSL (Ubuntu) from the start menu.
...
-want to build the defined development environment and use them. Installing (and
+want to build the defined development environment and use them. 
+
+Installing (and
 uninstalling) Nix is quite simple, thanks to the installer from [Determinate
 Systems](https://determinate.systems/posts/determinate-nix-installer), a company
-that provides services and tools built on Nix. Simply open a terminal and run
+that provides services and tools built on Nix.
+
+Instead of using any system package manager, simply open a terminal and run
```
b-rodrigues, line 100: Could you rewrite like this: "Do not use your operating system’s package manager to install Nix. Instead, simply open a terminal and run ..."

b-rodrigues (changes requested):
Thanks, there’s only a couple points to change

jrosell:
Suggestions applied. I insist in the 'Use' or 'Using' part of the title change.

b-rodrigues:
thanks for the contribution!

PR #295, opened 2024-09-12
Polish lint and styling actions

philipp-baumann:


.github/workflows/format-code.yaml
```diff
@@ -0,0 +1,25 @@
...
+  pull_request:
+    branches: [main]
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  main:
+    runs-on: ubuntu-22.04
```
b-rodrigues, line 12: Can we run this on 24.04?

philipp-baumann: yeah, think this should be possible. Will add that once working.

philipp-baumann: now added latest tag

philipp-baumann:
avoid race conditions: like implemented in this PR: https://github.com/ccao-data/model-res-avm/pull/244

philipp-baumann:
- [ ] avoid race conditions: like implemented in this PR: [ccao-data/model-res-avm#244](https://github.com/ccao-data/model-res-avm/pull/244)

philipp-baumann:
take home/lessions learnt:

- styler action: tweak styling in this action so that the R setup is bootstrapped via nix and our root default.nix with the necessary dev packages. adapt https://github.com/r-lib/httr2/blob/main/.github/workflows/pr-commands.yaml ; we now have auto-fixing in the PR, only if we have changes. 
- combine styling and linting. for linting, {lintr} is used. the styler action can write and auto-commit fixes, however the linter is not an auto-formatter. styler does very basic styling, and has caveats, and unfortunately has no auto-fixing of the line length implemented.

- I disabled

Issue #293, opened 2024-09-12
Alphabetically sort rPackages and pkgs in other categories in generated `default.nix`


philipp-baumann:
better version control and overview

PR #296, opened 2024-09-15
sorting packages

b-rodrigues:
Addresses #293


b-rodrigues:
should we go all the way and sort git packages as well? the way I see it, we would need a function that sorts a list on the basis of the "package_name" element. What do you think @philipp-baumann ?

b-rodrigues:
for reference

https://stackoverflow.com/a/28100627

PR #306, opened 2024-09-18
fixing tests

b-rodrigues:


philipp-baumann:
> using withr seems to be the way to go: it's easy to create local tempdirs that are used by all the underlying functions and it's easy to cleanup. Still need to deal with the following vignettes:
> 
>     * using nix inside docker
> 
>     * binary cache
> 
>     * bleeding edge
> 
>     * contributing to nixpkgs
> 
> 
> I think we might also want to use it for the functions that write to disk, like `hash_url()`

This really helps with the diagnostics, thanks a lot! i believe we should really not add another dependency, that is already known for depreciations. The solution is as simple as that in base R and we can just use the patterns in `withr::local_tempdir()` and `withr::local_tempfile`.

https://github.com/r-lib/withr/blob/d6b7d875a3a352a3667ff106ab29ddb9adc0caa0/R/tempfile.R#L63-L66

philipp-baumann:
should be all fixed by https://github.com/ropensci/rix/pull/308 -> we can close this one

PR #314, opened 2024-09-19
`R CMD check`/vignettes: fix `tempdir()`unlink top-level tempdir too

philipp-baumann:


philipp-baumann:
giving up, rhub checks are still failing, vignettes not compiling

PR #321, opened 2024-09-20
Fix `R CMD build` and `R CMD check` in standard `default.nix` dev env

philipp-baumann:


philipp-baumann:
no more "bleeding_edge". Whatever the reason is, there were permission issues.

```sh
nix-shell --pure
# change to rix pkg folder
R CMD build .
R CMD check ../rix_0.11.0.tar.gz
```

If bleeding edge nix env, then we get tempdir() permission errors. Checks always pass on "latest", and the vignettes build.

philipp-baumann:
@b-rodrigues We should have the common practice that we update the dev env for new CRAN releases, or maybe minor releases on GitHub. I would really manually control the env in nix with fixed SHA256-checksum pinning and make sure we can run `R CMD check` and `R CMD build` in a nix environment.

b-rodrigues:
I'm still not entirely convinced to be honest, ideally I would prefer my dev env to reflect the status on CRAN as much as possible. But let's see how this goes, we don't have that many dependencies anyways.

Issue #5, opened 2023-08-01
bootstrap helper from renv.lock


philipp-baumann:
Hey @b-rodrigues , I thought it would be nice for people to be able to transition from a {renv} env to a nix setup in individual projects. For this, a helper that simply parses JSON in `renv.lock` and grabs the packages for the `rix::rix()` `r_pkgs` arg would be straight-forward to implement. If useful, I am happy to draft a PR addressing it.

b-rodrigues:
that would be an amazing addition please feel free to code it up 😁

philipp-baumann:
maybe I will wait until you commit your recent drafts into new remote branch :-)

b-rodrigues:
I think a challenge here is to first list only the required packages that are actually called by the user using `library(xxx)` and not all of `xxx` dependencies, because these would need to got into propagatedBuildInputs. Packages would need to get downloaded from the CRAN archives. We can use fetchzip() to do so: 
https://github.com/b-rodrigues/rix/blob/d505382ad8ef759a8257227a19043e228a6d70dd/R/find_rev.R#L142

If we have the list of packages that are actually called by the user, and pass that to fetchzip() (or rather, fetchzips() which handles multiple packages see: https://github.com/b-rodrigues/rix/blob/d505382ad8ef759a8257227a19043e228a6d70dd/R/find_rev.R#L203

then that should be simple enough I think.

philipp-baumann:
I'll also have a look at how {renv} implements the snapshotting.

philipp-baumann:
directionality: renv => nix

RichardJActon:
I implemented the 'dumbest' version of this in my fork for personal use generating default.nix files from projects with renv.lock files. For the couple of projects with big renv.lock files I've tried it out on I've had to delete any packages installed from github from the list as these, expectedly, cause build errors. Also the `gt` package which seems to be failing because of it's dependency of V8. 

https://github.com/RichardJActon/rix/commit/706127425f10a38155cd1d9785641d2790b2f313

This may be something for it's own issue but here are the logs for the V8 build issue:
```
building '/nix/store/w4n5szzkapj921gfgf5yynijdvc2r15r-r-V8-5.0.0.drv'...
Running phase: unpackPhase
unpacking source archive /nix/store/92nvrcn6w8k2ylbrjzmd3lalwri0c4mv-V8_5.0.0.tar.gz
source root is V8
[...]

PR #356, opened 2024-11-03
helper functions to bootstrap from renv.lock 2

RichardJActon:
Resuming where we accidentally left off: https://github.com/ropensci/rix/pull/349

Partially addresses https://github.com/ropensci/rix/issues/5, this implements a "fast" inexact conversion from renv to nix environment description. It does **NOT** capture the exact package versions, just the R version and the list of packages in `renv.lock`, nor does it differentiate primary from transitive dependencies in `renv.lock` it simply lists all dependencies.

## To Do

- [ ] Finish testing on real world `renv.lock` examples, and handling any resulting edge cases
- [ ] Do any updates need to be made to the README / other docs about this?


RichardJActon:
About excluding {renv} from `default.nix` (https://github.com/ropensci/rix/pull/349#issuecomment-2450366574), I'm not sure this is always a good idea.

I've been playing around with projects where both the {renv} and nix environment management are present and running an `renv::snapshot()` then using the resulting `renv.lock` to generate a `default.nix` via `renv2nix()` is quite handy to do on a local environment managed with nix.
This is especially the case when my remote system is still doing builds in CI based on the {renv} tooling and not nix yet, this lets me keep both up to date and approximately in sync.

I'd say if the {renv} package is in your `renv.lock` then it should end up in your `default.nix`. It's worth noting that {renv} does not necessarily end up in the lock file by default when you use {renv} to manage a project. I just happen to have picked a couple of examples where it is in there on purpose as I extract the version of {renv} I'm using in that project from the lock file to install the right version in the container my CI uses.

b-rodrigues:
Thanks for reopening @RichardJActon and sorry again! Next time I want to add changes, I'll open a PR against your branch !

b-rodrigues:
We can ignore the unit test not passing in CI, that's due to a known issue with Rcpp and R 4.4.2. A fix is coming to CRAN so this won't be an issue much longer

b-rodrigues:
Looks like I was able to push without destroying the PR unlike last time :)
I’ll try to add some changes and merge by the end of the week

b-rodrigues:
As discussed on matrix, we now merge this gather already some feedback before spending more time writing tests. Many thanks to @RichardJActon for this much requested feature!

PR #359, opened 2024-11-18
Switch from using RemoteType to RemoteHost to fetch git packages in renv2nix

b-rodrigues:


b-rodrigues:
Currently this deals with #363

Issue #371, opened 2024-12-02
rix Error in if (grepl("\\.tar\\.gz", path)) { : the condition has length > 1


mihem:
While `renv2nix` worked fine on my previous project, `renv2nix` now fails with 
```
Error in if (grepl("\\.tar\\.gz", path)) { : the condition has length > 1
```

However, I realized that this is not a renv2nix issue, but an issue with rix in general.

Gist with the renv.lock https://gist.github.com/mihem/1cba71fe8df4980b1952a0d54479e245

@b-rodrigues I think this is a really great and complex renv-lock. You can also find the resulting rix-call in the traceback:

Traceback

```
15: get_imports(desc_path)
14: hash_url(url)
13: hash_git(repo_url = repo_url, commit)
12: nix_hash(repo_url, commit)
11: get_sri_hash_deps(repo_url, commit)
10: FUN(X[[i]], ...)
9: lapply(git_pkgs, fetchgit)
8: paste(lapply(git_pkgs, fetchgit), collapse = "\n")
7: fetchgits(git_pkgs)
6: paste(fetchgits(git_pkgs), fetchzips(archive_pkgs), collapse = "\n")
5: fetchpkgs(git_pkgs, archive_pkgs)
4: sprintf("\n  git_archive_pkgs = [%s  ];\n", fetchpkgs(git_pkgs, 
       archive_pkgs))
3: generate_git_archived_pkgs(git_pkgs, cran_pkgs$archive_pkgs, 
       flag_git_archive)
2: paste(generate_header(nix_repo, r_ver, rix_call), generate_rpkgs(cran_pkgs$rPackages, 
       flag_rpkgs), generate_git_archived_pkgs(git_pkgs, cran_pkgs$archive_pkgs, 
       flag_git_archive), generate_tex_pkgs(tex_pkgs), generate_local_r_pkgs(local_r_pkgs, 
       flag_local_r_pkgs), generate_system_pkgs(system_pkgs, r_pkgs), 
       generate_wrapped_pkgs(ide, attrib, flag_git_archive, flag_rpkgs, 
           flag_local_r_pkgs), generate_shell(flag_git_archive, 
           flag_rpkgs, flag_tex_pkgs, flag_local_r_pkgs, flag_wrapper, 
           shell_hook), collapse = "\n")
1: rix(r_ver = "4.3.1", r_pkgs = c("AnnotationDbi", "AnnotationFilter", 
[...]

PR #372, opened 2024-12-02
try fixing getting desc_path

mihem:
https://github.com/ropensci/rix/issues/371

This fixes the problem that occurs when there are two Description files resulting in  two paths.
I tried to solve it in a rather generic way, by using `file.path(list.files(path_to_src), "DESCRIPTION")` so only allowing the description file if it's in the root folder. This solves the problem in my use-case, but  @b-rodrigues feel free to change and further test.

### DIFF
--- a/R/nix_hash.R
+++ b/R/nix_hash.R
@@ -92,7 +92,7 @@ hash_url <- function(url) {
   sri_hash <- nix_sri_hash(path = path_to_source_root)
 
   paths <- list.files(path_to_src, full.names = TRUE, recursive = TRUE)
-  desc_path <- grep("DESCRIPTION", paths, value = TRUE)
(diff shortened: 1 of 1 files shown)


R/nix_hash.R
```diff
@@ -92,7 +92,7 @@ hash_url <- function(url) {
   sri_hash <- nix_sri_hash(path = path_to_source_root)
 
   paths <- list.files(path_to_src, full.names = TRUE, recursive = TRUE)
```
b-rodrigues, line 94: I wonder if it wouldn’t be enough to remove `recursive = TRUE` instead could you try this?

mihem: Not that s not enough because the path is hidden in the tmp folder, so just removing recursive leads to no match.

b-rodrigues:
many thanks!

PR #382, opened 2025-01-04
Handle remote deps

b-rodrigues:
This PR allows rix to handle packages with remote dependencies. For now, only remote that are hosted on Github are correctly handled, and only if they are written as `Remotes: r-lib/testthat` in the main package's description file. Remote dependencies of remote dependencies are also handled. 

TODO:

- [x] Update unit tests
- [x] Handling tags, hashes or PR (for example `r-lib/testthat@c67018fa4970`
- [x] To make tests stable, I need to fork lookup's deps and deps of deps as well

fyi @mihem


b-rodrigues:
@mihem where you able to build an env with your package scMisc? I see it also comes with a remote dep

mihem:
Well I used my CSFAtlasTool which worked great, but there was an easy when calling the remote dependency package `datathin` or using `renv2nix`. So I tried to fix this https://github.com/ropensci/rix/pull/385

PR #396, opened 2025-01-25
Fix rm vscode tmp

mihem:
@b-rodrigues this fixes a bug i had since the beginning in vscode after running rix and took me some time to figure out.

```
Warning message:
In file(con, "w") :
  cannot open file '/tmp/RtmpRf8FWl/vscode-R/workspace.json': No such file or directory
```

So vscode-R saves tmp files in tempdir

```
print(list.files(tempdir(), recursive = TRUE))
[1] "vscode-R/addins.json"    "vscode-R/response.lock" 
[3] "vscode-R/response.log"   "vscode-R/workspace.json"
[5] "vscode-R/workspace.lock"
```

rix deletes tmp files that are created (which is good i think) but also deletes all other tmp files, e.g. these vscode-R files, which produce these warning in vscode-R.

I tried to fix this by creating a rix_tmp specific folder.
At least in my case this fixes this issue

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: rix
 Title: Reproducible Data Science Environments with 'Nix'
-Version: 0.14.6
+Version: 0.14.7
 Authors@R: c(
     person(given = "Bruno", family = "Rodrigues", email = "bruno@brodrigues.co", role = c("aut", "cre"),
            comment = c(ORCID = "0000-0002-3211-3689")),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 <!-- NEWS.md is maintained by https://cynkra.github.io/fledge, do not edit -->
 
+# rix 0.14.7 (2025-01-26)
+
+- `rix()`: fixed a bug where temp files necessary to VS Code were being removed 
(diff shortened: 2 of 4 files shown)


b-rodrigues:
So if I understand correctly, the vscode-r extension also uses the same tempdir as the R session? So when you start a session in VS Code and generate a default.nix expression you get that error message?

mihem:
Yeah, vscode-R seems to also use tempdir. So I you start a R session in VS Code everything works fine. But after you run rix, those temporary files that vscode-R needs are deleted because of rix. So as long as the session continues, you get thar error message. You can just create a new R session and it's gone. But when I tried to debugging the get_cloest_commit stuf, this got really annoying because it was showing up all the time.
In general I would also think that it's dangerous if rix just deleted all temporary files? So it's safer to create a separate folder and only remove this folder when rix is finished.

b-rodrigues:
At first we actually had a separate temp folder but had many issues with some unit tests and trying to rebuild vignettes. I think that now it should be fine, though, as the vignettes are static now.

b-rodrigues:
closing in favour of https://github.com/ropensci/rix/pull/403#issuecomment-2629071522

PR #398, opened 2025-01-27
Remote namespace

mihem:
@b-rodrigues This tries to implement our idea to also include imports from NAMESPACE; not only from DESCRIPTION.

This was actually easier than I thought. I might have missed something, but the tests seems to pass.
I added tests, one checks if `Seurat` is fetched in `seurat-data`, which we added manually before that because it's not listed in DESCRIPTION.

I fixed a bug which happens because e.g. this CRAN mirror uses quotes. 
https://github.com/cran/AER/blob/master/NAMESPACE

Excited about your revision.

### DIFF
--- a/R/fetchers.R
+++ b/R/fetchers.R
@@ -274,6 +274,17 @@ get_imports <- function(path) {
   # Remove minimum package version for example 'packagename ( > 1.0.0)'
   output <- trimws(gsub("\\(.*?\\)", "", output))
 
+  # Get imports from NAMESPACE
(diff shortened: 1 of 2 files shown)


mihem:
Maybe there's one remaining issue with remote package. So if a remote package is called in NAMESPACE, but not in remotes, then this will not work because I think the username is not stated in NAMESPACE? However, I haven't found an example where a remote dependency is stated in NAMESPACE. Even in Seurat-Wrappers with those many dependency that's not the case https://github.com/satijalab/seurat-wrappers/blob/master/NAMESPACE. And i don't think there's anything we can do about this?

PR #399, opened 2025-01-31
Fix r profile

mihem:
@b-rodrigues I found that the resulting .Rprofile has some curly braces which prevent functions to work.

```
{
    is_rstudio <- Sys.getenv("RSTUDIO") == "1"
...
    rm(is_rstudio, is_nix_r)
}
```

So I could call `install.packages()` and there was not stop message.
Getting rid of the curly braces was quite difficult. Tried using expression or using subsitute but that did not work. So in the end i justed used gsub to remove them. Maybe not very elegant, but at least works. I tested that in my case it raises the expected error message. Not sure we can create a unit test for that, because it's testing the resulting environment?

Additionally, vscode-R did not work properly, `.vsc.attach` did not work and functionalities like `View` did not work.
https://github.com/REditorSupport/vscode-R/issues/1094

So I added the necessary source lines. I put this in an if statement only evaluating if the editor is vscode, so I think this should not mess up anything.

Looking forward to your feedback.

### DIFF
--- a/R/rix_init.R
+++ b/R/rix_init.R
@@ -138,9 +138,10 @@ rix_init <- function(project_path,
   # first create the call, deparse it, and write it to .Rprofile
   rprofile_quoted <- nix_rprofile()
   rprofile_deparsed <- deparse_chr1(expr = rprofile_quoted, collapse = "\n")
+  rprofile_cleaned <- gsub(pattern = "^\\{|\\}$", replacement = "", x = rprofile_deparsed)
   rprofile_file <- file.path(project_path, ".Rprofile")
 
-  rprofile_text <- get_rprofile_text(rprofile_deparsed)
+  rprofile_text <- get_rprofile_text(rprofile_cleaned)
 
   # This function creates the connection, write the text
   # and closes the connection
@@ -418,7 +419,7 @@ set_nix_path <- function() {
 #' @noRd
 nix_rprofile <- function() {
   # nolint start: object_name_linter
-  quote(
+  quote({
     is_rstudio <- Sys.getenv("RSTUDIO") == "1"
     is_nix_r <- nzchar(Sys.getenv("NIX_STORE"))
     is_code <- Sys.getenv("TERM_PROGRAM") == "vscode"
@@ -483,10 +484,10 @@ nix_rprofile <- function() {
       rm(current_paths, userlib_paths, user_dir, new_paths)
     }
 
-    rm(is_rstudio, is_nix_r)
     if (isTRUE(is_code) && interactive() && isFALSE(is_rstudio)) {
         source(file.path(Sys.getenv(if (.Platform$OS.type == "windows") "USERPROFILE" else "HOME"), ".vscode-R", "init.R"))
     }
-  )
+    rm(is_rstudio, is_nix_r, is_code)
+  })
   # nolint end: object_name
 }
--- a/R/rix_init.R
+++ b/R/rix_init.R
@@ -418,9 +418,10 @@ set_nix_path <- function() {
 #' @noRd
 nix_rprofile <- function() {
   # nolint start: object_name_linter
-  quote({
+  quote(
     is_rstudio <- Sys.getenv("RSTUDIO") == "1"
     is_nix_r <- nzchar(Sys.getenv("NIX_STORE"))
+    is_code <- Sys.getenv("TERM_PROGRAM") == "vscode"
     if (isFALSE(is_nix_r) && isTRUE(is_rstudio)) {
       # Currently, RStudio does not propagate environmental variables defined in
       # `$HOME/.zshrc`, `$HOME/.bashrc` and alike. This is workaround to
@@ -483,6 +484,9 @@ nix_rprofile <- function() {
     }
 
     rm(is_rstudio, is_nix_r)
-  })
+    if (isTRUE(is_code) && interactive() && isFALSE(is_rstudio)) {
+        source(file.path(Sys.getenv(if (.Platform$OS.type == "windows") "USERPROFILE" else "HOME"), ".vscode-R", "init.R"))
+    }
+  )
   # nolint end: object_name
 }


b-rodrigues:
Could you give more details? I just tried running install.packages("dplyr") inside of a Nix session, and I do get the error message. Same for update.packages().

mihem:
Hm, this is weird. On my machine (ubuntu noble) this did not raise the expected error. I tried making it reproducible using docker (and then encountered the issue https://github.com/ropensci/rix/issues/401 ), but the docker image then raised the expected error. So i need to investigate further why this happens with my settings in a non-docker environment.

b-rodrigues:
could you post the generated .Rprofile here?

mihem:
I couldn't reproduce my earlier problems; sorry, don't know what happened there. So I removed the gsub part which removed the curly braces and only kept the part for vscode-R.  I also added a check for positron because this is also "vscode" but does work with vscode-R. https://positron.posit.co/extensions.html

b-rodrigues:
where you starting the session with `--vanilla` perhaps?

mihem:
No. Well, everything works fine now, so I am happy :P.
Is it possible to create unit tests to check if the .RProfile is properly setup, so raises the expected errors in the nix-environment?

mihem:
@b-rodrigues any further input you need from me for this PR and the other PRs?

b-rodrigues:
can you explain what the Positron check does ?

mihem:
See https://positron.posit.co/extensions.html
If it s positron, this should not get called.

mihem:
image


So for vscode-R this is not important, but I think if you use Positron adding 
`source(file.path(Sys.getenv(if (.Platform$OS.type == "windows") "USERPROFILE" else "HOME"), ".vscode-R", "init.R"))`
breaks things, that's why I included the Positron check.

b-rodrigues:
I need to know exactly what this fixes before merging though

b-rodrigues:
From my reading this seems to be something rix shouldn't "fix"

mihem:
With the current .Rprofile, vscode-R is not really usable.

E.g. `View(iris)` opens a new windows, which looks terrible:

image

I guess you know this from using emacs.

while after this "fix" this looks as it's supposed to be in vscode-R

image

So adding this one line is just something which vscode-R users need to do to get vscode-R working properly: https://github.com/REditorSupport/vscode-R/wiki/R-Session-watcher#advanced-usage-for-self-managed-r-sessions

This is only one additional line, Rstudio has also specific settings and only if your editor is vscode, but not Positron, and if your session is interactive, this will be used. This will then source this file:
https://github.com/REditorSupport/vscode-R/blob/master/R/session/init.R

b-rodrigues:
I see but isn't vs code-r not mandatory with Positron? it seems to me that if someone wants to use vscode-r with Positron then it's on them to apply the suggested fix from the documentation, no?

b-rodrigues:
Ultimately, this has nothing to do with Nix nor rix, or did I misunderstand something? it is possible to append the .Rprofile instead of overwriting it so these users would need to append

mihem:
I only use vscode-R (tried Positron, but Copilot does not work, cannot use spacemacs keybindings using VSpaceCode as i am used to, devcontainer and ssh experimental)
No, from my understanding Positron is incompatible with vscode-R. 
So personally right now, I am only interested in using vscode-R, which require this line. You could say I have to manually add this line for each rix project, but of course it would be convenient if works out of the box, and I don't see any disadvantage here.

I just included checking if it's Positron because otherwise also Positron would source this file, which would not work if you read the Positron docs.

mihem:
Well, maybe it's not directly linked to nix or rix, but you also have special seetings for Rstudio or vscode (installing language server) and you create a specific R profile.
So as long as I am not in the nix-shell everaything works fine, but when I use the nix-shell vscode-R functionality does not work because of the custom .Rprofile that nix/rix uses. So I think rix should take care of getting back this functionality.

b-rodrigues:
OK I see I was confusing Positron and vscode

b-rodrigues:
I'll be testing and merging all PR this weekend hopefully

mihem:
Thanks!
I would think that all PRs except the get closest commit should be trivial, but of course the devil is in the details, so I am happy if you test this.

b-rodrigues:
Could you describe the issue step by step? Because I can't replicate. I have installed vscodium-fhs from nixpkgs, and trying to plot seems to work fine:

image

this is with an .Rprofile generated from the code in `main` and no `{httpgd}`. It's with this R extension

image


that's the one, right?

mihem:
Yes this is the right extension and everything works fine, which you can see because your plot looks good but also your terminal is attached to the code (becase it says R 4.4.2).  And you have an "R interactive" session, which I am not able to create using nix shell.

However, I think you don't use R in a nix-shell? Or are you opening vscode from within nix-shell?

I am using my locally installed VSCode. 

So I open my locally installed VSCode and when I then open a R terminal liek this (or via VSpacecode with Spc-m-m)

image

I get an i`Rinteractive`  terminal, where everything works fine, however this is my local R and not nix.

image


If I want to use R in nix I open a shell (zsh in my case) and run "nix-shell" and then "R"

image

THen you can see the terminal is called `R` not `R interactive`. And in the bottom row `R: (not attached)`.

If i click on this `R: (not attched), i get the following error (since the init.R of vscode-R is not sourced). 

image

And `View(iris` opens this terrible window shown above.

b-rodrigues:
i open vscode from a running nix-shell, what happens if you do that ?

b-rodrigues:
if you don't want to start vscode from a nix shell then this extension might help https://marketplace.visualstudio.com/items?itemName=arrterian.nix-env-selector

mihem:
I haven't thought about these options thanks. However, both options had some issues (or probably the same issue):

So using the extensions you proposed:
It's attached to the editor, which is great. However the libPaths is wrong (I think the first one shouldn't be in there at all and then why is there only one nix path in there), and I can't load my library then.

image


Using vscode from within nix-shell very much the same:

image

b-rodrigues:
are you inside of nested shells?

mihem:
Out of curiosity: How do handle this, e.g. in Emacs?

And for my understanding: What is the disadvantage of changing the Rprofile? At least I don't really know how to fix this with libPaths. And using nix-shell and R from within vscode is just a self-managed  R terminal described int he documentation and similar to tmux, so don't think this is such a hacky solution

image

b-rodrigues:
A priori no disadvantage but I want to make sure I understand what's wrong. When I start vscode from a shell everything works as expected so I need to know what's about your setup that breaks things because it could be the symptom of a more general issue.
Changing the .Rprofile could potentially break things for other setups.

Rstudio is the only editor that must be managed by Nix to work with Nix shell, but other could be locally managed.

As for me with emacs, I have emacs installed from openSuse, but use nix-direnv https://github.com/nix-community/nix-direnv

so when I start an R session in a folder with a nix-shell it gets loaded automatically

mihem:
Hm this could easily happen because I would use nnn for directory navigation and then change to shell again.
So I now didn't use nnn to get to that directory.

Using the extension I didn't know why there should be a nested shell, I think it's not

image

Still same issue then:
image

mihem:
And the other options: 

zsh is level1 and nix-shell is then level2.

image


But then .. tada everything works.

image

b-rodrigues:
can you share the .Rprofile and default.nix files here ?

mihem:
Usually I like to stay within the editor (earlier emacs,now vscode), and use extensions (projects) to navigate to my project and then open files and the R terminal from there (can use Spacemacs keybindings). 
I think the option of navigating to the folder using the shell (i have to remember not to use nested shell which can easily happen with my file manager nnn),  running nix-shell and opening code from there is okay, maybe a little less preferable from my view.

I would still prefer adding this PR to make it work with self-managed R terminals for vscode users (I can't see why this would break things for others because it will only apply if you use vscode), and we could change it if this does actually break the setup for other users .

b-rodrigues:
And if you don't have a system wide library of packages, this works as well ?

mihem:
thanks for sharing your workflow (althouth I am not fully understanding nix-direnv).

And to sum up:
You are right, vscode works fine when started locally. However vscodeR needs this init.R script if it's not started within nix-shell.

mihem:
Ah, now I think I am finally understand the problem you have with the approach.
You mean if you install everything using nix and don't have system libraries at all. I don't think this will apply to many R users, but sure would be nice if it would also work then.
I am not sure there is a way I can test this. Isn't this your setup? Can you test this?
I mean we could also include another statement within the if vscode statement. If the file exists then use this, otherwise skip this and maybe print a message that this has to be done manually or link to the docs (where we should better explian that the editor has to be started from within nix-shell)

b-rodrigues:
> Can you test this?

I'd be quicker for you to rename the R library in your home/ so it doesn't get loaded than for me to setup vscode and extensions and whatnot 😉

mihem:
Ah sorry, I though you mean having everything installed in /nix. 
So i renamed R and everything works fine. 
If i rename `.vscode-R` then i get an error message of course
```
Error in file(filename, "r", encoding = encoding) : 
  cannot open the connection
In addition: Warning message:
In file(filename, "r", encoding = encoding) :
  cannot open file '/home/mischko/.vscode-R/init.R': No such file or directory
```

And as before when the code of this PR wasn't included I cannot attach the R terminal

```
> .vsc.attach()
Error in .vsc.attach() : could not find function ".vsc.attach"
```

mihem:
So I added a chech if the file exists (so this would be apply if someone uses VSCode interactively (not Positron), but does not have vscode-R installed (which is unlikely because you cannot do much without an R extension), but still. I hope this is what you meant?

b-rodrigues:
I'll be doing some final testing and then this should be good to merge.

It also made me think that the vignette titled "interactive use" needs to be updated and describe all of this as well.

b-rodrigues:
So I installed vs code from the opensuse package manager and opened a project with a default.nix and got this message

image

I already had these two extensions (perhaps from previous testing?)

image

Starting the command palette and selecting "Nix-Env: select environment" and then restarting vs code works. Then, using `View(mtcars)` opens the data in that awful window as you described.

I then tried with the code in your branch, but got following warning message:

```
Warning messages:
1: In rm(is_rstudio, is_nix_r, is_code, is_positron) :
  object 'is_rstudio' not found
2: In rm(is_rstudio, is_nix_r, is_code, is_positron) :
  object 'is_nix_r' not found
3: In rm(is_rstudio, is_nix_r, is_code, is_positron) :
  object 'is_code' not found
4: In rm(is_rstudio, is_nix_r, is_code, is_positron) :
  object 'is_positron' not found
```

and the .Rprofile is just `NULL`. Can you replicate? Try in a completely empty directory. Here is the call to `rix()` I used:

```
library(rix)

rix(
  date = "2025-01-27",
  r_pkgs = c("httpgd", "dplyr", "ggplot2"),
  ide = "code",
  project_path = ".",
  overwrite = TRUE,
  print = TRUE
)
```

mihem:
@b-rodrigues sorry, my bad. It worked before. But with the last commit, I put the last two parentheses at the wrong place and didn't test anymore.

I tested it now and it worked for me. Here is the resulting .Rprofile

```
### File generated by `rix::rix_init()` ###
# 1. Currently, system RStudio does not inherit environmental variables
#   defined in `$HOME/.zshrc`, `$HOME/.bashrc` and alike. This is workaround to
#   make the path of the nix store and hence basic nix commands available
#   in an RStudio session
# 2. For nix-R session, remove `R_LIBS_USER`, system's R user library.`.
#   This guarantees no user libraries from the system are loaded and only
#   R packages in the Nix store are used. This makes Nix-R behave in pure manner
#   at run-time.
{
    is_rstudio <- Sys.getenv("RSTUDIO") == "1"
    is_nix_r <- nzchar(Sys.getenv("NIX_STORE"))
    is_code <- Sys.getenv("TERM_PROGRAM") == "vscode"
    is_positron <- Sys.getenv("POSITRON") == "1"
    if (isFALSE(is_nix_r) && isTRUE(is_rstudio)) {
        cat("{rix} detected RStudio R session")
        old_path <- Sys.getenv("PATH")
        nix_path <- "/nix/var/nix/profiles/default/bin"
        has_nix_path <- any(grepl(nix_path, old_path))
        if (isFALSE(has_nix_path)) {
            Sys.setenv(PATH = paste(old_path, nix_path, sep = ":"))
        }
        rm(old_path, nix_path)
    }
    if (isTRUE(is_nix_r)) {
        install.packages <- function(...) {
            stop("You are currently in an R session running from Nix.\n", "Don't install packages using install.packages(),\nadd them to ", "the default.nix file instead.")
        }
        update.packages <- function(...) {
            stop("You are currently in an R session running from Nix.\n", "Don't update packages using update.packages(),\n", "generate a new default.nix with a more recent version of R. ", "If you need bleeding edge packages, read the", "'Understanding the rPackages set release cycle and using ", "bleeding edge packages' vignette.")
        }
        remove.packages <- function(...) {
            stop("You are currently in an R session running from Nix.\n", "Don't remove packages using `remove.packages()``,\ndelete them ", "from the default.nix file instead.")
        }
        current_paths <- .libPaths()
        userlib_paths <- Sys.getenv("R_LIBS_USER")
        user_dir <- grep(paste(userlib_paths, collapse = "|"), current_paths, fixed = TRUE)
        new_paths <- current_paths[-user_dir]
        .libPaths(new_paths)
        rm(current_paths, userlib_paths, user_dir, new_paths)
    }
    if (isTRUE(is_code) && interactive() && isFALSE(is_rstudio) && isFALSE(is_positron)) {
        vscode_r_init <- file.path(Sys.getenv(if (.Platform$OS.type == "windows") 
            "USERPROFILE"
        else "HOME"), ".vscode-R", "init.R")
        if (file.exists(vscode_r_init)) {
            source(vscode_r_init)
        }
        else {
            message("No .vscode-R/init.R file found. If you want to use VSCode-R, you need to source it in your .Rprofile or start vscode from within nix-shell")
        }
    }
    rm(is_rstudio, is_nix_r, is_code, is_positron)
}
```

b-rodrigues:
nice, I'll do some more testing and will likely merge tomorrow

mihem:
I also made a short screenrecording to show how I would use it

https://github.com/user-attachments/assets/c341e143-c5b0-4446-910f-2289e204a8c5


I renamed ~/.vscode-R (so testing what would happen if you don't have vscode-R extension installed) and then it shows the message and it's not attached

image

b-rodrigues:
whatever does this mean ? not attached ?

mihem:
In contrast, I played around with starting code from nix-shell and somestimes encountered difficulties. So if i already had an instance of vscode running, then the new vscode, which was started from within nix-shell, showed again this one nix location only and one of my system library.

image


So I am more happy from starting vscode without nix-shell.

mihem:
@b-rodrigues `not attached` just means that the R-terminal is not attached to the file. That's because vscode-R is not working properly, so you cannot use all the goodies and have this awful View etc.

R/rix_init.R
```diff
@@ -481,8 +482,16 @@ nix_rprofile <- function() {
...
     }
-
-    rm(is_rstudio, is_nix_r)
+    # source vscode-R init.R file for vscode-R
+    if (isTRUE(is_code) && interactive() && isFALSE(is_rstudio) && isFALSE(is_positron)) {
+      vscode_r_init <- file.path(Sys.getenv(if (.Platform$OS.type == "windows") "USERPROFILE" else "HOME"), ".vscode-R", "init.R")
+      if (file.exists(vscode_r_init)) {
+        source(vscode_r_init)
+      } else {
+        message("No .vscode-R/init.R file found. If you want to use VSCode-R, you need to source it in your .Rprofile or start vscode from within nix-shell")
```
b-rodrigues, line 491: Is the extension called VSCode-R? Is it this one? https://open-vsx.org/extension/REditorSupport/r

Shouldn't it be REditorSupport?

mihem: well it's often called vscode-R and it's stored in the location `.vscode-R`.
Here the github repo also says vscode-R https://github.com/REditorSupport/vscode-R.
As far as I know it's the only (or at least only official) R extension in vscode

b-rodrigues:
So this seems to be working fine, but I think that the best workflow likely is:

- install [direnv](https://direnv.net/docs/installation.html)
- create an `.envrc` file in the root folder of your project (next to the `default.nix` file) and allow following content: `use nix`
- Install the vcode direnv extension: https://marketplace.visualstudio.com/items?itemName=mkhl.direnv
- Restart vscode. You'll have to allow direnv (you'll get a notification). You might also need to restart the R shell.
- Now VSCode automatically loads the shell

This is actually exactly how I do it with Emacs. Just put `.envrc` file in all your projects and allow them (only need to do this once) and VSCode will seamlessly switch envs.

This is ready to merge, there's just one comment left for you to address.

Many thanks for this contribution!

R/rix_init.R
b-rodrigues: fair enough

mihem:
Thank you.

I will definitely try this out. And VSCode will then also activate the nix-environment? That would save two steps (but requires an extra package, and extensions and creating a file).

b-rodrigues:
> I will definitely try this out. And VSCode will then also activate the nix-environment? 

Yep, once you install direnv, create an .envrc file and allow direnv in that directory, vs code will switch automatically and transparently.

mihem:
I've tried this and it worked great.
I agree this is also a great option, maybe even slightly better because you don't have to manually activate the shell and start R.

b-rodrigues:
I’ll be documenting this: https://github.com/ropensci/rix/issues/405

Issue #401, opened 2025-02-01
dockerfile docs outdated?


mihem:
Hi @b-rodrigues,

when trying to use Dockerfile as described in docs right now https://docs.ropensci.org/rix/articles/z-advanced-topic-using-nix-inside-docker.html
docker build start compiling R, which takes forever.
When I used a different `default.nix` as a starting point, this does not happen.
So  comment out old version, use newer version (which you once provided in a GitHub issue)
```
#RUN curl -O https://raw.githubusercontent.com/ropensci/rix/main/inst/extdata/default.nix
RUN curl -O https://gist.githubusercontent.com/b-rodrigues/162679cc0ca131181ec140d8a340886d/raw/09526bc89a93ef83e23a661dd99a12b18316c068/default.nix
```

Is this because of an outdated rstats-on-nix archive?

b-rodrigues:
could you try again? This should be pulling the latest available date which should be in the cache now.

mihem:
Hm, I would also have guessed that this fixes things, but strangely this then leads again to R compiling.

[docker_log_v3.log](https://github.com/user-attachments/files/18628581/docker_log_v3.log)

mihem:
starting from `#11 98s`

b-rodrigues:
see this one is cached

```
#7 [ 4/10] RUN curl -O https://raw.githubusercontent.com/ropensci/rix/main/inst/extdata/default.nix
#7 CACHED
```

so it's using the old default.nix still

also are you setting up the cache using `setup_cachix()`? I don't see it, see the dockerfile from the vignette there's a step that does it

mihem:
Yes I just using the Dockerfile from your vignette. You can see in the log file step 7:

`#11 [ 7/10] RUN nix-shell --run "Rscript -e 'rix::setup_cachix()'"`

But you are right, docker is automatically using the cached version. I used `--no-cached` and now it works fine.
So you are right I used the old version before and with your latest PR you fixed things as expected.

PR #402, opened 2025-02-01
Update make_inst_default.yaml

b-rodrigues:
deals with #401 

Probably what's happening is that the R version used is now too old, and got purged from the cache. Let's try with a more recent date.

fyi @mihem


b-rodrigues:
working on both linux and macOS

PR #403, opened 2025-02-01
Remove excessive unlinks

b-rodrigues:
I wonder if instead of #396 we shouldn't be instead removing all of this. It's ok to keep files in the temporary directories, what's not ok is:

- writing to the user directory and keeping the files there
- tests or vignettes that write to the user directory and keep the detritus there

I think that this approach would be saner.

### DIFF
--- a/R/fetchers.R
+++ b/R/fetchers.R
@@ -217,27 +217,26 @@ get_imports <- function(path) {
     # to only keep packagename
     remotes <- gsub("\n", "", x = unlist(strsplit(remotes$Remotes, ",")))
     # Get user names
-    remote_pkgs_usernames <- strsplit(remotes, "/") |>
-      sapply(function(x) x[[1]])
+    remote_pkgs_usernames <- sapply(strsplit(remotes, "/"), function(x) x[[1]])
 
     # Now remove user name and
     # split at "@" or "#" character to get name and commit or PR separated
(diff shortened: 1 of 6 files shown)


mihem:
Yeah, also thought about that but I didn't want to change too much code. Also I wasn't sure if CRAN or so were complaining if there were temporary files left.
I would think both approaches are fine.

mihem:
> Packages should not write in the user’s home filespace (including clipboards), nor anywhere else on the file system apart from the R session’s temporary directory (or during installation in the location pointed to by TMPDIR: and such usage should be cleaned up).

from: https://cran.r-project.org/web/packages/policies.html

This sounds as if it was actually better to clean up the files?

b-rodrigues:
>apart from the R session’s temporary directory (or during installation in the location

that's what `tempdir()` does, no? I'm trying to build the package on win-builder.r-project.org to see but the service seems to be down

mihem:
Hm okay, maybe you are right, then you only clean up the files that are creatied during installation, but not necessarily the files created during calling a function? I would think deleting all temporary files created by the function is cleaner, but yeah probably they are cleaned up anyway and it's not strictly forbidded by CRAN or so.

b-rodrigues:
@mihem could you please test this PR? For cran this is okay

mihem:
I tested a rix call and similar to my PR this solves the problem of removing the vscode-R files stored in tmp directory.

PR #411, opened 2025-02-06
Better ide handling

b-rodrigues:
Ned to update the vignette


mihem:
Sorry, just saw this by accident.
If i understand correctly, these change would mean that if I select `code` that it will install the nixpkg of `code`. 
I think the solution you showed me (using local vscode+direnv extension, direnv) works great. Adding nixpkg of code adds quite some overload. You could say I should just choose `none`, however I think without languageserver this is not really useable. So shouldn't the default be just languageserver, and then people can decide for themselves if they also want to install the nixpkgs vscode (the same for other IDEs)?

b-rodrigues:
you're right this is still something I'm thinking about how to handle best. Perhaps with another option to select whether the IDE should be installed by nix or not? I don't know yet what the best interface should be

mihem:
Hm I would think it should be as minimal as possible. I can really only speak for vscode. `languageserver` is always required so it makes sense to always have this added.
I think it's fine as it is right now. You can tell users in the docs that they can install their IDE by adding `vscode-fhs` to system pkgs.

b-rodrigues:
I also like minimalist approaches in general, but I think that offering an option to also add the editor easily is helpful, especially because of the config.allowUnfree option that must be set for Code and Positron. So I’m thinking add an argument, something like `install_ide = FALSE` on top could work? So if you write `ide = "code", install_ide = FALSE` you only get `languageserver` but if you set it to TRUE it also adds vscode-fhs and the required option. Same for other editors. By default, it could stay FALSE. 

This could also be useful for (Neo)-Vim which also requires some specific configuration, see https://github.com/R-nvim/R.nvim/issues/229#issuecomment-2642408464

what do you think?

b-rodrigues:
🤔 on the other hand, why not do it the other way around? if `ide = some_ide` then it’s going to be installed by Nix automatically. However, if you want to use your system-level VS Code, why not add languageserver to r_pkgs yourself, and set ide to none? that might be more consistent, no?

b-rodrigues:
> 🤔 on the other hand, why not do it the other way around? if `ide = some_ide` then it’s going to be installed by Nix automatically. However, if you want to use your system-level VS Code, why not add languageserver to r_pkgs yourself, and set ide to none? that might be more consistent, no?

I think that’s what I’m doing 😆 for users like you, it’s just a matter of writing `r_pkgs = "languageserver"` instead of `ide = "code"`. Also leaves the door open to use another LSP if possible.

mihem:
Yeah, agree that's also fine by me. But then ide should default to `none`.

Issue #416, opened 2025-02-10
Problems when generating the default.nix file with rix


zecojls:
Hi, I just want to report an issue that I solved but might be useful for you.

I used rix to generate the project `default.nix` of my project but I had to manually change so many things inside to be able to build the environment. In the end, and after so much trial and error (which I cannot reproduce anymore), it worked. I remember that some package dependencies were being called more than once that Nix was flagging as duplicates and failing the build. Also, some dependencies failed that I had to relocated or indicate them in the main call before installing the GitHub repositories.

You can find what is working for me on this [default.nix file](https://github.com/soilspectroscopy/ossl-nix/blob/main/default.nix).

This is the [rix code](https://github.com/soilspectroscopy/ossl-nix/blob/main/files/rix_generate_environment.R) I used to generate the template that needed to be modified to work.

I think you can apply a diff to see what has changed.

My session (PS: I use [rig](https://github.com/r-lib/rig) to manage multiple versions of R, in this case, my default is 4.3.3).
```
R version 4.3.3 (2024-02-29)
Platform: aarch64-apple-darwin20 (64-bit)
Running under: macOS 15.3

Matrix products: default
BLAS:   /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib 
LAPACK: /Library/Frameworks/R.framework/Versions/4.3-arm64/Resources/lib/libRlapack.dylib;  LAPACK version 3.11.0

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

time zone: America/New_York
tzcode source: internal

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] rix_0.14.3

loaded via a namespace (and not attached):
[1] compiler_4.3.3    tools_4.3.3       rstudioapi_0.17.1 curl_6.0.1        codetools_0.2-20 
[6] jsonlite_1.8.9   
```

And thank you for developing reproducible research tools/tutorials like rix.

b-rodrigues:
Thank you for reporting this! I think that this is an issue that is now solved in the development version of rix, I'll try your rix call with the latest venison tomorrow and report back. If it is still available broken, I'll work on a fix!

b-rodrigues:
Ok, so there’s several things:

- first of all, the duplicated entries should be a thing of the past with the development version of rix now
[...]

Issue #100, opened 2024-01-19
Remember to link the vignette in the other vignette


b-rodrigues:
Don't forget it bruno!

see: https://b-rodrigues.github.io/rix/articles/building-reproducible-development-environments-with-rix.html

Issue #11, opened 2023-08-03
{fusen} blows an error


philipp-baumann:
```r
> fusen::inflate_all()
✔ The flat file build_envs.Rmd is going to be inflated
✔ The flat file building_envs_with_rix.Rmd is going to be inflated
✔ The flat file cran_archive.Rmd is going to be inflated
✔ The flat file data_doc.Rmd is going to be inflated
✔ The flat file interactive_use.Rmd is going to be inflated
✔ The flat file non_interactive_use.Rmd is going to be inflated
✔ The flat file save_r_nix_revs.Rmd is going to be inflated
✔ The flat file zzz.Rmd is going to be inflated
Error in parse(text = code) : <text>:43:0: unexpected end of input
41: 
42: 
   ^
> fusen::inflate_all()
✔ The flat file build_envs.Rmd is going to be inflated
✔ The flat file building_envs_with_rix.Rmd is going to be inflated
✔ The flat file cran_archive.Rmd is going to be inflated
✔ The flat file data_doc.Rmd is going to be inflated
✔ The flat file interactive_use.Rmd is going to be inflated
✔ The flat file non_interactive_use.Rmd is going to be inflated
✔ The flat file save_r_nix_revs.Rmd is going to be inflated
✔ The flat file zzz.Rmd is going to be inflated
Error in parse(text = code) : <text>:43:0: unexpected end of input
41: 
42: 
   ^
> traceback()
15: parse(text = code)
14: is.factor(x)
13: gsub("\\n", " ", as.character(parse(text = code)))
12: parse_fun(rmd_fun[x, ])
11: FUN(X[[i]], ...)
10: lapply(seq_len(nrow(rmd_fun)), function(x) parse_fun(rmd_fun[x, 
        ]))
9: get_functions_tests(parsed_tbl)
8: (function (pkg = ".", flat_file, vignette_name = "Get started", 
       open_vignette = TRUE, check = TRUE, document = TRUE, overwrite = "ask", 
       ...) 
   {
       if (!is.null(list(...)[["name"]])) {
           stop(paste0("The `name` argument to `inflate()` is deprecated since {fusen} version 0.3.0.", 
               "\nPlease use `vignette_name = '", list(...)[["name"]], 
               "'` instead.\n"))
           vignette_name <- list(...)[["name"]]
       }
       if (!is.null(list(...)[["rmd"]])) {
[...]

PR #420, opened 2025-02-12
Handle GitHub tag pr

mihem:
@b-rodrigues tries to address some of the problems learned from https://github.com/ropensci/rix/issues/416

If there is a remote dependency with a PR (so e.g. `pkg#27`) this will be removed (so it will then call resolve_package_dependencies to get the closeset commit)

Tags work out of the box.

Added tests for those situations.


mihem:
Okay maybe I was a little to optimistic.
The tests all pass (both locally and also on github), but they don't test if then `nix-shell` also works.

If i run `nix-shell` with the expression I get from 

```
rix(
  date = "2025-02-10",
  r_pkgs = NULL,
  system_pkgs = NULL,
  git_pkgs = list(
      package_name = "rix",
      repo_url = "https://github.com/ropensci/rix/",
      commit = "v0.8.0"
  ),
  ide = "none",
  project_path = ".",
  overwrite = TRUE,
  print = TRUE
)
```

which is

```
let
 pkgs = import (fetchTarball "https://github.com/rstats-on-nix/nixpkgs/archive/2025-02-10.tar.gz") {};
  
    rix = (pkgs.rPackages.buildRPackage {
      name = "rix";
      src = pkgs.fetchgit {
        url = "https://github.com/ropensci/rix/";
        rev = "v0.8.0";
        sha256 = "sha256-E4WYQeQRPuIKPZY7TEudcSW9AxNc0KDKs7+QV2U7sjI=";
      };
      propagatedBuildInputs = builtins.attrValues {
        inherit (pkgs.rPackages) 
          codetools
          curl
          jsonlite
          sys;
      };
    });
    
  system_packages = builtins.attrValues {
    inherit (pkgs) 
      glibcLocales
      nix
      R;
  };
  
in

pkgs.mkShell {
  LOCALE_ARCHIVE = if pkgs.system == "x86_64-linux" then "${pkgs.glibcLocales}/lib/locale/locale-archive" else "";
  LANG = "en_US.UTF-8";
   LC_ALL = "en_US.UTF-8";
   LC_TIME = "en_US.UTF-8";
   LC_MONETARY = "en_US.UTF-8";
   LC_PAPER = "en_US.UTF-8";
   LC_MEASUREMENT = "en_US.UTF-8";

  buildInputs = [ rix   system_packages   ];
  
}
```

I get


```
warning: error: unable to download 'https://api.github.com/repos/NixOS/nixpkgs/commits/nixpkgs-unstable': HTTP error 403

       response body:

       {"message":"API rate limit exceeded for 37.201.156.150. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}

```

Actually the building step works fine and I can activate my shell and then get rix v0.8.0 as requested. However, I am little confused by this error message. Does that mean that GithubAPI is then used during the building process? This is probably not want we want? And why does this succeed despite the warning? @b-rodrigues

mihem:
Hm this tag thing is more complicated than I thought. I think the big difference is that before I was resolving the package dependencies in R and tried to get a commit, which has a close date and then nix works fine. Here, I thought that it's fine to just use the tag.  I though this because that's what you did with your strsplit function (tried to get the ref) and also using try_download with a tag instead of has works fine

```
url1 <- "https://github.com/mlr-org/mlr3extralearners/archive/6e2af9ef9ecd420d2be44e9aa2488772bb9f7080.tar.gz"
url2 <- "https://github.com/mlr-org/mlr3extralearners/archive/v1.0.0.tar.gz"
path <- tempfile()
handle <- curl::new_handle(failonerror = TRUE, followlocation = TRUE)
req1 <- curl::curl_fetch_disk(url1, path = path, handle = handle)
req2 <- curl::curl_fetch_disk(url2, path = path, handle = handle)
```
However, I now realize that try_download working is only one part (to get the correct imports), but then also the resulting default.nix has to work with nix (and this is not tested with the tests in nix).

As I said in my previous post that worked fine on my local machine. However,  to be on the safe side I tried to run the same rix call in a Dockerfille, so

generate_env.R

```
library(rix)
rix(
  date = "2025-02-10",
  r_pkgs = NULL,
  system_pkgs = NULL,
  git_pkgs = list(
      package_name = "rixTest",
      repo_url = "https://github.com/mihem/rixTest",
      commit = "25da90697895b006934a70bbd003aab5c5206c8b"
  ),
  ide = "none",
  project_path = ".",
  overwrite = TRUE,
  print = TRUE
)

```


Don't think i have seen these thousands
`#11 31.89 warning: Nix search path entry '/root/.nix-defexpr/channels' does not exist, ignoring` 
before.

[docker_v1.log](https://github.com/user-attachments/files/18774991/docker_v1.log)

In the end this build is successful, and I also get the expected version. 

image


However I also get these strange warnings agains

image


A

b-rodrigues:
> As I said in my previous post that worked fine on my local machine. However, to be on the safe side I tried to run the same rix call in a Dockerfille, so

The way this is done here is through dedicated actions like this one: https://github.com/ropensci/rix/blob/main/.github/workflows/test-renv2nix.yaml

An expression gets generated, and then built. If it fails, we know the generated expression is wrong on some level. This actually really helped me yesterday, when I was working on the other PR. Could you add an action like this (with as many steps as needed to build as many default.nix files as needed) to this PR?

> Don't think i have seen these thousands `#11 31.89 warning: Nix search path entry '/root/.nix-defexpr/channels' does not exist, ignoring` before.
> 
> [docker_v1.log](https://github.com/user-attachments/files/18774991/docker_v1.log)

I’ve seen these as well, gotta admit I have no idea what it is.

> So to sum up @b-rodrigues
> 
> 1. I think the PR thing is solved (well actually just ignored, but works better this way)
> 2. I am still unsure about tags. I think it would be nice to get this working and it actually seems to work, but maybe you can double check
> 3. there are some new warnings I get when using rix in a dockerfile (although everything build successfuly in the end and works fine)

1. just add some actions for integration testing and I’ll be reviewing the code later in more detail
2. How are branches handled? For example here, the remotes point to a branch https://github.com/tidymodels/workflows/blob/1c38711d28ae3ea27bc3d240d68bbd44a032a113/DESCRIPTION which doesn’t exist anymore. Anyway to distinguish branches from tags?
3. yeah I think we can ignore those

mihem:
Thanks, that's exactly what I meant.

I started to setup a github workflow for PR, but I don't think we need that. Because since rix now removes the PR, it will use download_all_commit and then get the commit hash, and we don't need to test that this works.
So I now set up a workflow for a tag https://github.com/ropensci/rix/actions/runs/13316582419/job/37191975157?pr=420
I actually skipped the step of calling rix to create the default nix (because the relevant part is already tested here https://github.com/ropensci/rix/blob/0aed1a0c2ab274aae02e906bb7fc95db9ee83947/tests/testthat/test-fetchers.R#L137)

and this seems to work fine :)

2) good point, will work on that

3) yeah it works, but I found these a little annoying, there are also present in the github workflows, you are right, but I am sure that a few weeks ago they were not present, see e.g. https://github.com/ropensci/rix/actions/runs/13101533215/job/36550373602

mihem:
3) this was last week ago and still no errors: https://github.com/ropensci/rix/actions/runs/13125152572/job/36619860375
Not sure this is the right way, but I went to the latest PR and I saw it for the first the in the better IDE handling PR checks

https://github.com/ropensci/rix/actions/runs/13260868909/job/37016963753

b-rodrigues:
Probably an update to the action install nix, setting up some option or something that doesn't make sense on GA?

mihem:
3) Hm not sure I can help there

2) Hm damn it. You are right, when using a branch name nixpkgs cannot work with that.

So using a tag name, rix can just pass it to nix and this works fine

```
    rix = (pkgs.rPackages.buildRPackage {
      name = "rix";
      src = pkgs.fetchgit {
        url = "https://github.com/ropensci/rix";
        rev = "v0.8.0";
        sha256 = "sha256-E4WYQeQRPuIKPZY7TEudcSW9AxNc0KDKs7+QV2U7sjI=";
      };
      propagatedBuildInputs = builtins.attrValues {
        inherit (pkgs.rPackages) 
          codetools
          curl
          jsonlite
          sys;
      };
    });

```

but if using a branch name like this, it does not work.

```
     rix = (pkgs.rPackages.buildRPackage {
      name = "rix";
      src = pkgs.fetchgit {
        url = "https://github.com/ropensci/rix";
        rev = "available_dates";
        sha256 = "sha256-qaILF0MugXBY/rwFph8m1FHoLQUT8ym0KVaM+cLlP3Y=";
      };
      propagatedBuildInputs = builtins.attrValues {
        inherit (pkgs.rPackages) 
          codetools
          curl
          jsonlite
          sys;
      };
    });
```

The problem is that both are separated by a `@` and there are no rules for tag or branch names.
So I think this is impossible to distinguish.  Only option I see would be to use another github API to fetch this information. But actually I agree with you that this makes sense very complex and that this are rather edge cases.

So I think i will just not only remove everything and including `#` but also `@`. We also don't need a new workflow then (also it was intersting to see how this works). This means that remote_pkgs_names_and_refs is then never a list of two.

mihem:
ah I forgot, we may want to keep it if it's a commit hash. However this should be doable with a regex.

mihem:
@b-rodrigues So i think I am done.

The way this is handled now.

Ignore PR (so `#`) ignore tags, branches, only take into account commit shas (`@` and then a regex `^[0-9a-f]{7,40}$`). I think chances are pretty low that someone would call his branch or tag this way.
I wrote test for PR and tag, in both situations a commit hash is fetched via the GitHub API. 
I used the github workflow to test if this works with commit sha also in the short form  (dummy example https://github.com/mihem/rixTest/commit/e2fc06592877bde459855364f9e7115e0cd6d409) and this worked fine.
I also updated the docs.

Looking forward to your revision

mihem:
Ah one last thing. In github workflow I had to use my fork for now, because it's not merged. After you merge this I think it's better if you would change the url, so this line in test-fetch-commit-short.yaml

```
wget -O default.nix https://raw.githubusercontent.com/mihem/rix/refs/heads/handle_github_tag_pr/tests/testthat/testdata/remote-pkgs/fetch_commit_short_default.nix
```

tests/testthat/test-fetchers.R
```diff
@@ -90,6 +90,96 @@ testthat::test_that("Test fetchgits", {
...
+      package_name = "rixTest",
+      repo_url = "https://github.com/mihem/rixTest",
+      commit = "25da90697895b006934a70bbd003aab5c5206c8b"
+    )
+  )
+  expected_output <- paste0(
+    "\n    rix = (pkgs.rPackages.buildRPackage {\n",
+    "      name = \"rix\";\n",
+    "      src = pkgs.fetchgit {\n",
+    "        url = \"https://github.com/ropensci/rix\";\n",
```
b-rodrigues, line 153: So if I understand correctly, rixTest has `Remotes: ropensci/rix#100` in its DESCRIPTION, but the PR simply gets ignored. However, the commit that gets used is then the closest to the one from rixTest, right?

b-rodrigues: That’s how it’s described in the vignette, so I guess that’s right

mihem: yes. PR, branch and tags are ignored (although tags would work with the nix expression, but cannot be differentaited from branches I think), only commit shas are used. I think this makes sense and nicely fits to rix philosophy of always trying to use a commit hash. So if a commit hash is given it's used, all others fixed tags/pr/branch are not used.

mihem: and yes if nothing is speciied then the resolve_package_commit is used to fetch the closest commit hash.

b-rodrigues:
I just tried to rerun this one:

```
rix(
  date = "2023-12-30",
  r_pkgs = c("tidyverse", "mlr3", "qs", "matrixStats", "prospectr", "Cubist",
             "checkmate", "mlr3misc", "paradox"),
  system_pkgs = NULL,
  git_pkgs = list(
    package_name = "mlr3extralearners",
    repo_url = "https://github.com/mlr-org/mlr3extralearners/",
    commit = "6e2af9ef9ecd420d2be44e9aa2488772bb9f7080"
  ),
  ide = "none",
  project_path = path_default_nix,
  overwrite = TRUE,
  print = TRUE
)

```
and got :

```
Error: Request `curl::curl_fetch_disk()` failed:
HTTP response code said error [codeload.github.com]: The requested URL returned error: 404
If it's a GitHub repo, check the url and commit.
Are these correct? If it's an archived CRAN package, check the name
of the package and the version number.
Failing repo: https://github.com/xoopR/distr6/archive/292.tar.gz
```

shouldn't this be ignored now?

mihem:
However, the default nix still has the duplicate, which you aimed to remove with your postprocessing that is already merged to main branch I think.

```
error: attribute 'set6' already defined at /home/mischko/develop/rix_test_new/default.nix:58:5
       at /home/mischko/develop/rix_test_new/default.nix:131:5:
          130|
          131|     set6 = (pkgs.rPackages.buildRPackage {
             |     ^
          132|       name = "set6";
```

b-rodrigues:
Argh, sometimes I should just go to bed :) I confirm everything works now post-processing was not being run because of a typo. All good now :)

mihem:
Ah that's good that it also works for you.

Yes, also still duplicates.
I think your regexp is not working

`grepl("^\\s*([a-zA-Z0-9_]+)\\s*=\\s*\\(", line)` was `FALSE`

and the line looks like this, so doesn't look like a line but more like everything?

```
"# This file was generated by the {rix} R package v0.15.1 on 2025-02-15\n# with following call:\n# >rix(date = \"2023-12-30\",\n#  > r_pkgs = c(\"tidyverse\",\n#  > \"mlr3\",\n#  > \"qs\",\n#  > \"matrixStats\",\n#  > \"prospectr\",\n#  > \"Cubist\",\n#  > \"checkmate\",\n#  > \"mlr3misc\",\n#  > \"paradox\"),\n#  > system_pkgs = NULL,\n#  > git_pkgs = list(package_name = \"mlr3extralearners\",\n#  > repo_url = \"https://github.com/mlr-org/mlr3extralearners/\",\n#  > commit = \"6e2af9ef9ecd420d2be44e9aa2488772bb9f7080\"),\n#  > ide = \"none\",\n#  > project_path = \".\",\n#  > overwrite = TRUE,\n#  > print = TRUE,\n#  > r_ver = \"4.3.2\")\n# It uses the `rstats-on-nix` fork of `nixpkgs` which provides improved\n# compatibility with older R versions and R packages for Linux/WSL and\n# Apple Silicon computers.\n# Report any issues to https://github.com/ropensci/rix\nlet\n pkgs = import (fetchTarball \"https://github.com/rstats-on-nix/nixpkgs/archive/2023-12-30.tar.gz\") {};\n \n  rpkgs = builtins.attrValues {\n    inherit (pkgs.rPackages) \n      checkmate\n      Cubist\n      matrixStats\n      mlr3\n      mlr3misc\n      paradox\n      prospectr\n      qs\n      tidyverse;\n  };\n \n    CoxBoost = (pkgs.rPackages.buildRPackage {\n      name = \"CoxBoost\";\n      src = pkgs.fetchgit {\n        url = \"https://github.com/binderh/CoxBoost\";\n        rev = \"1dc47d7051f660b28520670b34d031143f9eadfd\";\n        sha256 = \"sha256-5Ck0idCpn6oU3RBULqcz7bEtOnYxzKH29aHVzOUWGsQ=\";\n      };\n      propagatedBuildInputs = builtins.attrValues {\n        inherit (pkgs.rPackages) \n          survival\n          Matrix\n          prodlim;\n      };\n    });\n\n\n    set6 = (pkgs.rPackages.buildRPackage {\n      name = \"set6\";\n      src = pkgs.fetchgit {\n        url = \"https://github.com/xoopR/set6\";\n        rev = \"e65ffeea48d30d687482f6706d0cb43b16ba3919\";\n        sha256 = \"sha256-trJ2cmx/KXRapN+LoHbyGNBueHhLCDWvrtZSicJba5U=\";\n      };\n      propagatedBuildInputs = builtins.attrValues {\n        inherit (pkgs.rPackages) \n          checkmate\n          ooplah\n          Rcpp\n          R6;\n      };\n    });\n\n    param6 = (pkgs.rPackages.buildRPackage {\n      name = \"param6\";\n      src = pkgs.fetchgit {\n        url = \"https://github.com/xoopR/param6\";\n        rev = \"0fa35771276fc05efe007a71bda466ced1e4c5eb\";\n        sha256 = \"sha256-6mfOzx0DPGnKyXJPFm1V1qhsLCIHC26XW8q5jZ2gpAg=\";\n      };\n      propagatedBuildInputs = builtins.attrValues {\n        inherit (pkgs.rPackages) \n          checkmate\n          data_table\n          dictionar6\n          R6;\n      } ++ [ set6 ];\n    });\n\n    distr6 = (pkgs.rPackages.buildRPackage {\n      name = \"distr6\";\n      src = pkgs.fetchgit {\n        url = \"https://github.com/xoopR/distr6\";\n        rev = \"548ba956510cb1b5389901ad6a960d7db4d16f41\";\n        sha256 = \"sha256-jTaDwXzyWyWOOmtWuY/H/Ys/3pDm1hGFtSZX/zHwrDE=\";\n      };\n      propagatedBuildInputs = builtins.attrValues {\n        inherit (pkgs.rPackages) \n          checkmate\n          data_table\n          ooplah\n          R6\n          Rcpp;\n      } ++ [ set6 param6 ];\n    });\n\n    mlr3proba = (pkgs.rPackages.buildRPackage {\n      name = \"mlr3proba\";\n      src = pkgs.fetchgit {\n        url = \"https://github.com/mlr-org/mlr3proba\";\n        rev = \"c5bec7b9b0b73d3611e61882e7556404a6d9fb2e\";\n        sha256 = \"sha256-xGOluZgEAPqRWm0HbhCaSa+qmonEQoBexud2h3BeY58=\";\n      };\n      propagatedBuildInputs = builtins.attrValues {\n        inherit (pkgs.rPackages) \n          mlr3\n          checkmate\n          data_table\n          ggplot2\n          mlr3misc\n          mlr3viz\n          paradox\n          R6\n          Rcpp\n          survival\n          survivalmodels;\n      } ++ [ distr6 param6 set6 ];\n    });\n\n\n    set6 = (pkgs.rPackages.buildRPackage {\n      name = \"set6\";\n      src = pkgs.fetchgit {\n        url = \"https://github.com/xoopR/set6\";\n        rev = \"e65ffeea48d30d687482f6706d0cb43b16ba3919\";\n        sha256 = \"sha256-trJ2cmx/KXRapN+LoHbyGNBueHhLCDWvrtZSicJba5U=\";\n      };\n      propagatedBuildInputs = builtins.attrValues {\n        inherit (pkgs.rPackages) \n          checkmate\n          ooplah\n          Rcpp\n          R6;\n      };\n    });\n\n    param6 = (pkgs.rPackages.buildRPackage {\n      name = \"param6\";\n      src = pkgs.fetchgit {\n        url = \"https://github.com/xoopR/param6\";\n        rev = \"0fa35771276fc05efe007a71bda466ced1e4c5eb\";\n        sha256 = \"sha256-6mfOzx0DPGnKyXJPFm1V1qhsLCIHC26XW8q5jZ2gpAg=\";\n      };\n      propagatedBuildInputs = builtins.attrValues {\n        inherit (pkgs.rPackages) \n          checkmate\n          data_table\n          dictionar6\n          R6;\n      } ++ [ set6 ];\n    });\n\n    distr6 = (pkgs.rPackages.buildRPackage {\n      name = \"distr6\";\n      src = pkgs.fetchgit {\n        url = \"https://github.com/alan-turing-institute/distr6\";\n        rev = \"bb410f4557395846906d8dbcbca9f0f71fc15900\";\n        sha256 = \"sha256-i0zMfYZA4INu49J4RcKuxNczSXpSjkqdwXQZwbs0o/E=\";\n      };\n      propagatedBuildInputs = builtins.attrValues {\n        inherit (pkgs.rPackages) \n          checkmate\n          data_table\n          ooplah\n          R6\n          Rcpp;\n      } ++ [ set6 param6 ];\n    });\n\n    survivalmodels = (pkgs.rPackages.buildRPackage {\n      name = \"survivalmodels\";\n      src = pkgs.fetchgit {\n        url = \"https://github.com/RaphaelS1/survivalmodels\";\n        rev = \"9d59b0c93780a71ae8a6c9904eed72a360a1c2d6\";\n        sha256 = \"sha256-LvrT6425UU3gMY3xpXCz/iE9I1GH7gsWxTKOO8KMpVU=\";\n      };\n      propagatedBuildInputs = builtins.attrValues {\n        inherit (pkgs.rPackages) \n          Rcpp;\n      } ++ [ distr6 param6 set6 ];\n    });\n\n    mlr3extralearners = (pkgs.rPackages.buildRPackage {\n      name = \"mlr3extralearners\";\n      src = pkgs.fetchgit {\n        url = \"https://github.com/mlr-org/mlr3extralearners/\";\n        rev = \"6e2af9ef9ecd420d2be44e9aa2488772bb9f7080\";\n        sha256 = \"sha256-zdoZUSdL90uZcUF/5nxrNhZ9JdVRb6arstW1SjAACX8=\";\n      };\n      propagatedBuildInputs = builtins.attrValues {\n        inherit (pkgs.rPackages) \n          checkmate\n          data_table\n          mlr3\n          mlr3misc\n          paradox\n          R6;\n      } ++ [ distr6 CoxBoost mlr3proba survivalmodels param6 set6 ];\n    });\n    \n  system_packages = builtins.attrValues {\n    inherit (pkgs) \n      glibcLocales\n      nix\n      R;\n  };\n  \nin\n\npkgs.mkShell {\n  LOCALE_ARCHIVE = if pkgs.system == \"x86_64-linux\" then \"${pkgs.glibcLocales}/lib/locale/locale-archive\" else \"\";\n  LANG = \"en_US.UTF-8\";\n   LC_ALL = \"en_US.UTF-8\";\n   LC_TIME = \"en_US.UTF-8\";\n   LC_MONETARY = \"en_US.UTF-8\";\n   LC_PAPER = \"en_US.UTF-8\";\n   LC_MEASUREMENT = \"en_US.UTF-8\";\n\n  buildInputs = [ mlr3extralearners rpkgs  system_packages   ];\n  \n}"
```

btw i alsow watned to try to remove the duplicates, but it's good to have post processing  as a safety net

b-rodrigues:
that might be it, in the tests I pass a default.nix and it's read line by line, here it's passed as a single string which is why it's not working in integration testing

I'll fix ASAP

mihem:
yeah exactly, also just wanted to proprose 
`cat(strsplit(text, "\n")[[1]], sep = "\n")` but i think you've already done that

mihem:
@b-rodrigues yes that worked. set6 is now successfully removed.
I think it would make sense to adjust the test default.nix to be in one line form with `\n`.

Issue #430, opened 2025-02-27
Server not receiving information to compute hash


b-rodrigues:
The server is currently not receiving the right info required to compute the hash if nix is not present on the system. The right result is returned via curl though:

```
 curl -X GET "https://git2nixsha.dev/hash?repo_url=https%3A%2F%2Fgithub.com%2Feliocamp%2FmetR&commit=1dd5d391d5da6a80fde03301671aea5582643914" -H  "accept: */*"

```

but unit tests on a system without Nix fail:

```
── Error ('test-fetchers.R:131:3'): Test fetchgits works when PR is provided in a remote package, but does not use it ──
Error: Error: Request `curl::curl_fetch_memory(url = 'https://git2nixsha.dev/hash?repo_url=&commit=')` failed:
 HTTP response code said error [git2nixsha.dev]: The requested URL returned error: 500
If it's a GitHub repo, check the url and commit.
Are these correct? If it's an archived CRAN package, check the name
of the package and the version number.
```

see: https://github.com/ropensci/rix/actions/runs/13562423321/job/37911046377

fyi @mihem perhaps you see where this is coming from?

b-rodrigues:
calling fetchgit() individuall works fine

b-rodrigues:
Using the debugger:

- with the local nix install, repo_url is first CSFAtlasTools, and then get_sri_hash_deps gets called again and is datathin
- using the api server, when get_sri_hash_deps gets called again, this time repo_url is NULL

so we need to understand why that is

mihem:
In general, I think the issue is that we never checked nix_hash_online. I actually never changed code there but maybe indirectly broke something there or it never worked.

Yeah I found debugging really helpful, at work right now, so cannot do this (maybe this evening)

But yes that must be it, why is repo_url NULL.

mihem:
So when is repo_url NULL? when calling nix_hash_online?

b-rodrigues:
ok so the plot thickens; trying this directly with curl:

```
curl -X GET "https://git2nixsha.dev/hash?repo_url=https://github.com/mihem/CSFAtlasTools&commit=ac4a34d39812e8b8fa5746b63df4cf321a13b7a7" -H  "accept: */*"
```

results in the following:

```
{
    "sri_hash": ["sha256-w2yEeAMI/+zLgYeEfPRNzf25q35l0KSY8kiJXgeMBwI="],
    "deps": {
        "package": ["CSFAtlasTools"],
[...]

PR #431, opened 2025-02-27
Trying to fix response from the server

b-rodrigues:


b-rodrigues:
Dealing with #430 
this seems to be better than before, but it fallsback to HEAD

mihem:
Ah nice, yeah you are halfway there I think, and funny how this is just this never ending problem of R functions "simplifying" things, which leads to problems.

So now the only question is why getting the commit hash does not work. Some argument must be in the wrong format or so, your last commit seemed reasonable but still falling back to HEAD. I can debug this further this evening .. or you have further fun with this ;)

b-rodrigues:
yeah, unlisting `git_pkgs$commit$ doesn’t seem to solve the issue. 🤔

mihem:
Yeah I can see you are having fun with this ;)

b-rodrigues:
mmmh the server seems to crash randomly as well. I wonder if it's been down this whole time and no one noticed.... maybe I'll simply remove this feature altogether.

mihem:
Well I never really understood the use case pof not having nix installed.
What did you have in mind when you designed this?
It would certainly simplify things a lot because I think everything works fine as long as nix is installed. But you maybe had good reason to include this feature?

b-rodrigues:
> Well I never really understood the use case pof not having nix installed. What did you have in mind when you designed this? It would certainly simplify things a lot because I think everything works fine as long as nix is installed. But you maybe had good reason to include this feature?

the idea was to enable the generation of expressions on systems where nix is not included, for example to then commit on CI. Buuut one could argue that you could just as well generate the expression on CI and build it there, so...

mihem:
I would think that the use cases for most of the R users is to create reproducible pipelines of their local analysis on their system or using docker images. In both cases nix hash via server is not needed.
I think you are the one with the wildest use case (both for testing this package, creating your blog and so on).
So if you don't need it, I think no one else will miss is that soon. But definitely your call.

PR #441, opened 2025-03-01
fix typo docs

mihem:
@b-rodrigues typo I introduced


mihem:
ah you adjustd the docs a little .. but kept my typo :P

Issue #467, opened 2025-04-13
BPCells cannot be installed


mihem:
@b-rodrigues Hope you are doing fine.

I have quite some problems when trying to install GitHub package: https://github.com/bnprks/BPCells

**First issue:**
The package is pretty good and I use it quite a lot. However, it has this unusual structure "r" and "python" subdirectories, which causes quite some problems.

```
rix(
  date = "2025-04-11",
  r_pkgs = c("dplyr"),
  system_pkgs = NULL,
  git_pkgs = list(
   package_name = "BPCells",
   repo_url = "https://github.com/bnprks/BPCells/r",
   commit = "16faeade0a26b392637217b0caf5d7017c5bdf9b"
  ),
  ide = "none",
  project_path = ".",
  overwrite = TRUE,
  print = TRUE
)
```

fails becaus try_downoad fails:

```
Error: Request `curl::curl_fetch_disk()` failed:
HTTP response code said error [github.com]: The requested URL returned error: 404
If it's a GitHub repo, check the url and commit.
Are these correct? If it's an archived CRAN package, check the name
of the package and the version number.
Failing repo: https://github.com/bnprks/BPCells/r/archive/16faeade0a26b392637217b0caf5d7017c5bdf9b.tar.gz
```

That's because you can only download the entire repo no just the subdirectory R which we would need here.

So using the entire repo

```
rix(
  date = "2025-04-11",
  r_pkgs = c("dplyr"),
  system_pkgs = NULL,
  git_pkgs = list(
   package_name = "BPCells",
   repo_url = "https://github.com/bnprks/BPCells",
   commit = "16faeade0a26b392637217b0caf5d7017c5bdf9b"
  ),
  ide = "none",
  project_path = ".",
  overwrite = TRUE,
  print = TRUE
)
```

fails

```
Error in file(con, "r") : cannot open the connection
In addition: Warning message:
In file(con, "r") :
  cannot open file '/tmp/RtmpVJlmqN_repo_hash_url_xkcrw/file1d1354b0fbc6b/package_src/BPCells-16faeade0a26b392637217b0caf5d7017c5bdf9b/NAMESPACE': No such file or directory
```

When debugging, I found out that this fails because there's no NAMESPACE in that directory (actually code that I wrote I think :P).
There is some DESCRIPTION in the root folder, so the reading does not fail, but the DESCRIPTION in the root folder is useless.
We would need both files from subdirectory r.
We could easily achieve this in this case by modifying these lines to use the r folder.

```
[...]

PR #473, opened 2025-04-18
Remote subdirs

mihem:
@b-rodrigues as discussed here https://github.com/ropensci/rix/issues/467
rix should work with github/gitlab packages that use a subdirectory to store the R package like BPCells.

### DIFF
--- a/R/nix_hash.R
+++ b/R/nix_hash.R
@@ -24,13 +24,15 @@ nix_hash <- function(repo_url, commit, ...) {
 
 #' Return the SRI hash of an URL with .tar.gz
 #' @param url String with URL ending with `.tar.gz`
+#' @param repo_url URL to GitHub repository
+#' @param commit Commit hash
 #' @param ... Further arguments passed down to methods.
 #' @return list with following elements:
 #' - `sri_hash`: string with SRI hash of the NAR serialization of a GitHub repo
 #'      at a given deterministic git commit ID (SHA-1)
 #' - `deps`: list with three elements: 'package', its 'imports' and its 'remotes'
 #' @noRd
-hash_url <- function(url, ...) {
+hash_url <- function(url, repo_url, commit, ...) {
   tdir <- tempdir()
 
   tmpdir <- paste0(
@@ -65,15 +67,21 @@ hash_url <- function(url, ...) {
 
   tar_file <- file.path(path_to_tarfile, "package.tar.gz")
 
-  # url <- "https://github.com/bnprks/BPCells/r/archive/16faeade0a26b392637217b0caf5d7017c5bdf9b.tar.gz"
-  # root_url <- "https://github.com/bnprks/BPCells/archive/16faeade0a26b392637217b0caf5d7017c5bdf9b.tar.gz"
-
-
-  # root_url <- sub(
-  #   ".*github\\.com/[^/]+/[^/]+/(.+)/archive/.*",
-  #   "\\1",
-  #   url
-  # )
+  # remove subdirectory from URL if given because only the entire repo can be downloaded
+  if (unlist(strsplit(url, "/"))[6] != "archive") {
+    base_repo_url <- sub(
+      "(https://(github|gitlab)\\.com/[^/]+/[^/]+/).*",
+      "\\1",
+      repo_url
+    )
+    if (grepl("github", repo_url)) {
+      root_url <- paste0(base_repo_url, "archive/", commit, ".tar.gz")
+    } else if (grepl("gitlab", repo_url)) {
+      root_url <- paste0(base_repo_url, "-/archive/", commit, ".tar.gz")
+    }
+  } else {
+    root_url <- url
+  }
 
   try_download(
     url = root_url,
@@ -253,7 +261,7 @@ hash_git <- function(repo_url, commit, ...) {
     url <- paste0(repo_url, slash, "-/archive/", commit, ".tar.gz")
   }
   # list contains `sri_hash` and `deps` elements
-  hash_url(url, ...)
+  hash_url(url, repo_url, commit, ...)
 }
 
 
--- a/R/nix_hash.R
+++ b/R/nix_hash.R
@@ -65,8 +65,18 @@ hash_url <- function(url, ...) {
 
   tar_file <- file.path(path_to_tarfile, "package.tar.gz")
 
+  # url <- "https://github.com/bnprks/BPCells/r/archive/16faeade0a26b392637217b0caf5d7017c5bdf9b.tar.gz"
+  # root_url <- "https://github.com/bnprks/BPCells/archive/16faeade0a26b392637217b0caf5d7017c5bdf9b.tar.gz"
+
+
+  # root_url <- sub(
+  #   ".*github\\.com/[^/]+/[^/]+/(.+)/archive/.*",
+  #   "\\1",
+  #   url
+  # )
+
   try_download(
-    url = url,
+    url = root_url,
     file = tar_file,
     handle = h,
     extra_diagnostics = extra_diagnostics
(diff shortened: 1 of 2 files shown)


mihem:
Ah I did not pay attention to the fact that `hash_url` is not only used by github packages but also by cran packages, which don't have a repo_url or commit

mihem:
well actually it works with cran packages but not with archive packages because the they use hash_url, try to fix this

R/nix_hash.R
```diff
@@ -68,16 +68,20 @@ hash_url <- function(url, repo_url, commit, ...) {
...
-      "(https://(github|gitlab)\\.com/[^/]+/[^/]+/).*",
-      "\\1",
-      repo_url
-    )
-    if (grepl("github", repo_url)) {
-      root_url <- paste0(base_repo_url, "archive/", commit, ".tar.gz")
-    } else if (grepl("gitlab", repo_url)) {
-      root_url <- paste0(base_repo_url, "-/archive/", commit, ".tar.gz")
+  if (grepl("github|gitlab", url)) {
+    if (unlist(strsplit(url, "/"))[6] != "archive") {
```
b-rodrigues, line 72: how robust is selecting the 6th element?

mihem: I know what you mean, but I think it should be robust in this setting.
Because the user should always use
`https://github.com/bnprks/BPCells/r`

and it's always `/user/package`.

I actually used the same approach here

```
    repo_url_short <- paste(unlist(strsplit(url, "/"))[4:5], collapse = "/") 
```

But sure could also use regex, if you prefer that or if you can think of a scenario where it's not the 6th element.

mihem: ah okay, it does actually not work with gitlab because in gitlab you have `-/archive/` so it's the seventh elementh.
So okay i change it now to regex with "archive", should work as long as the package is not called archive :P

mihem: Pff, introduced quite some bugs by these changes, but I think I've fixed them all.

mihem:
So I think I am done with the PR, all tests pass. 
@b-rodrigues when you have time, please take a look

I think i will add two sentenced in the docs, will have time again on Sunday

R/nix_hash.R
```diff
@@ -82,18 +113,62 @@ hash_url <- function(url, ...) {
     path_to_src,
     list.files(path_to_src)
   )
-
   sri_hash <- nix_sri_hash(path = path_to_source_root)
 
-  paths <- list.files(path_to_src, full.names = TRUE, recursive = TRUE)
+  # extract subdirectory from GitHub or GitLab URL if given
+  # and create a path to this subdirectory
+  if (grepl("github", url)) {
```
b-rodrigues, line 120: this is a lot of repeated code for github and gitlub, could you move this to a function that generates the required code instead

R/nix_hash.R
```diff
@@ -65,8 +67,37 @@ hash_url <- function(url, ...) {
...
+        "(https://github\\.com/[^/]+/[^/]+/).*",
+        "\\1",
+        repo_url
+      )
+      root_url <- paste0(base_repo_url, "archive/", commit, ".tar.gz")
+    } else {
+      root_url <- url
+    }
+  } else if (grepl("gitlab", url)) {
+    has_subdir <- grepl("gitlab\\.com/[^/]+/[^/]+/[^/]+/-/archive/", url)
```
b-rodrigues, line 84: the regex patterns are repeated a lot, it might be better to put them in a function as well and return the right regex depending on the situation

R/nix_hash.R
```diff
@@ -82,18 +113,62 @@ hash_url <- function(url, ...) {
...
 
   if (grepl("github", url)) {
-    repo_url_short <- paste(unlist(strsplit(url, "/"))[4:5], collapse = "/")
+    repo_url_short <- sub(
+      "https://github\\.com/([^/]+/[^/]+)(/.*)?/archive/.*",
+      "\\1",
+      url
+    )
     commit <- gsub(x = basename(url), pattern = ".tar.gz", replacement = "")
     commit_date <- get_commit_date(repo_url_short, commit)
```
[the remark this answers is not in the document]
mihem: yeah, although this is more complicated than changing it here, because this relies on API github, github token etc, so yes maybe something for a future PR although I think we have seen several users relying on github remotes and not any relying on gitlab, also don't have any package from gitlab i use.

mihem:
@b-rodrigues I've created a function to generate the regex (with some help by Claude Sonnet), should be less redundant now, also created some docs, hope you are happy with that.

mihem:
btw found naother prominent example mentioned on the remotes github site.
https://github.com/dmlc/xgboost

R/nix_hash.R
```diff
@@ -65,8 +126,57 @@ hash_url <- function(url, ...) {
 
   tar_file <- file.path(path_to_tarfile, "package.tar.gz")
 
+  # First check which platform the URL is for
+  is_github <- grepl("github", url)
+  is_gitlab <- grepl("gitlab", url)
+
+  # Get the root URL based on platform
+  if (is_github) {
```
b-rodrigues, line 134: can't we do `if (has_subdir)` only once, and then in that if ... else do another if for is_github, and then is_gitlab? This should reduce the amount of code

b-rodrigues: I even wonder if we need `is_github` and is `is_gitlab` couldn't we simply set a variable say `platform` conditionally, and then use it everywhere with `get_git_regex(platform)` below?

b-rodrigues (changes requested):

b-rodrigues:
> btw found naother prominent example mentioned on the remotes github site. https://github.com/dmlc/xgboost

xgboost is on cran, so it's available through nixpkgs as well

mihem:
> > btw found naother prominent example mentioned on the remotes github site. https://github.com/dmlc/xgboost
> 
> xgboost is on cran, so it's available through nixpkgs as well

Yeah sure, but if you want to install the dev version (as mentioned on remotes here https://github.com/r-lib/remotes) you need a package which supports subfolder, which rix currently does not

R/nix_hash.R
mihem: Well I think that is what sonnet proposed however the problem is that this will fail if it not GitHub or giitlab, so e.g cran archive. Then get_git_regex wil be called, but will fail.

b-rodrigues: there's certainly a way to handle gitlab and Github with less code and do the exception for the CRAN archives

R/nix_hash.R
```diff
@@ -82,22 +175,42 @@ hash_url <- function(url, ...) {
     path_to_src,
     list.files(path_to_src)
   )
-
   sri_hash <- nix_sri_hash(path = path_to_source_root)
 
-  paths <- list.files(path_to_src, full.names = TRUE, recursive = TRUE)
+  # Process Git URL to handle subdirectories
+  if (!is.null(platform)) {
```
b-rodrigues, line 181: why is the if statement being repeated here? couldn’t you put the `url_subdir` and `path_to_r` variables already above, in the first if else statement?

mihem: sorry, what do you mean with repeated if statement here?
if it's not github/gitlab this does not need to be executed and we are done and we can set path_to_r. If it's github/gitlab we get the patterns and check if it has a subdir. If no subdir, we can skip the subdir stuff and set path_to_ro, but still neeed to construct the repo_url_short and get the commit and commit date (which is not necessary if it's just a cran archive; besides this only works for github, so we need to check if the platform is github, if you mean that with if statement being repeated?)
If it has a subdir then we also need to get the subdir.

If we would put the url_subdir and path_to_r n the first if statement, we could also execute it if it has no subdir, which is not necessary.

b-rodrigues: there’s two if statements to see if the platform is null and then if has_subdir (so starting on lines 139 and then again on line 181), couldn’t this be put into one? I don’t see why this should be tested twice. It seems to me that we could only have one, no?

mihem: Ah now I understand what you mean. You mean put the entire if statement of 181 ff into the if statement of 139 ff, right?
The first if statement (139 ff) is required for try_download (so that we download the root url and not the subdir which will fail) and the second if statement 181 ff is required to get the correct desc path (so that we use the subdir, where the desc is located).

Yeah sure, I could put this into one if statement I think, probably have time for this tomorrow.

mihem:
@b-rodrigues okay i put it in the first if statement, was some additional work and code.
I think it's cleaner now, and everything works fine. There is only on tests that fails regarding the lookup package from your repo.
I just corrected the output because I thought just the order was different but i now saw that i no longer has memoise and gh as remote output, need to investigate further why this is.

b-rodrigues:
you should revert that last commit, as the test is not fixed

mihem:
yeah I know, I try to understand why .. not sure yet, do you see it?

b-rodrigues (approved):

b-rodrigues:
Many thanks!

mihem:
Welcome

Issue #527, opened 2025-08-16
coverage of python packages


mihem:
@b-rodrigues 
It's great that rix also covers python packages.
I used the example in the vignettes, which worked great.

I then tested a few popular python package in my scientific field, e.g https://github.com/scverse/scanpy or https://github.com/scverse/scvi-tools.

Unfortunately (and surprisingly) python coverage really seems to be much worse on nixpkgs than for R, or more positive: you R maintainer really do a great job.
Actually, I saw that at least for scanpy there  is a PR on nixpkgs (more than a year after the discussion started https://github.com/NixOS/nixpkgs/issues/313389), but e.g. not for scvi-tools.

Do you have any idea how to extend this?

BTW: I tried pixi (which is super fast, very easy to set up, great user experience, works well with Python packages, but R packages don't work well since R-packages on conda-forge are outdated https://github.com/prefix-dev/pixi/issues/1543#issuecomment-3193917616

b-rodrigues:
Just as an aside, it should be noted that Julia for Nix has a similar approach than R for Nix, and if you look at the contributing guidelines for contributing Julia packages, you'll also notice some similarities with CRAN: https://github.com/JuliaRegistries/General?tab=readme-ov-file#faq

mihem:
Thanks for the info.

Hm okay, the way you describe it sounds not very optimistic.

But I mean as a data scientist you basically have to work in R or (or more often and) python. At least in my field it's about 60% R and 40% python, I mostly try to use R package (e.g. scanpy is the python equivalent to Seurat), but then sometimes there are methods only available in python (e.g. https://github.com/theislab/scvelo).

And then there is reproducibility in data science, which you wrote a entire book about. So how would you ensure reproducibility if you have R + python packages, and python packages not as common as polars (but still quite popular, ~500 stars, Nature Methods paper).

conda actually covers most (or I think so far all of them) of these packages, e.g. https://prefix.dev/channels/conda-forge/packages/scvelo.

From my point of view it seems that rix is great for reproducing R packages, but covers much too few python packages for using it in real projects (at least in my field), which also use python.

Isn't there a way to make use of conda for nix/rix?
It seems to be quite easy in pixi to build recipes https://github.com/wolfv/r-forge.

Side note: Yeah your ambitions for T are great. But if you want to use currently published packages (at least in my field) you have to use R or Python.

mihem:
[...]

PR #529, opened 2025-08-18
Docs python install uv

mihem:
@b-rodrigues see https://github.com/ropensci/rix/issues/527


b-rodrigues:
many thanks for raising this and providing a solution! I think that this could help further adoption of rix!

mihem:
You are welcome.

Agree, this means that users can install theoretically all python packages, while the amount of packages was quite limited before

Issue #557, opened 2025-11-03
Deduplication test not passing


b-rodrigues:
https://github.com/ropensci/rix/actions/workflows/test-dedup.yaml

it seems like instead of using the provided commit for one of the deps it falls back to HEAD, hence the snapshot test fails.

@mihem since you wrote the code do you mind taking a look? I won't have capacity for some time

mihem:
@b-rodrigues Yeah I remember that those two were always a pain to debug. On the other hand they are very useful because we had this duplication problem for quite some time. Also quite busy at the moment, bul I'll try to find time to debug this.

b-rodrigues:
nice that's likely it! Just to confirm could you try to fetch 2000 commits and see if the test passes? that would confirm this is indeed the case

mihem:
solved with https://github.com/ropensci/rix/pull/562

PR #562, opened 2025-11-09
Adjust max commits

mihem:
Increase max commits to 3000 to address errors, which occures in deduplication tests.
https://github.com/ropensci/rix/issues/557


b-rodrigues:
very nice , thank you!

mihem:
you are welcome, very easy fix, also happy that this works that easily.

So I also thought about your idea. I think we could leave it as it is, as everything works now.
But to make it more robust for the future, you are righe, we (or probably you, with your account)  could also fork all repos that are used in the tests and refer to this. Also not sooo much work.
This is what you proposed right?

Issue #559, opened 2025-11-06
Rix doesn't allow usage of GITHUB_PAT for accessing private repositories.


JonathanMatthewSamson:
When running rix::rix and trying to use a private repository, github will return 404 without use of a PAT.

b-rodrigues:
What do you mean you don't have permission to open a PR? You should be able to fork the repository, push your changes, and then open a PR, no?

JonathanMatthewSamson:
That might be my inexperience with forks, I usually push a branch and make a PR from that.  I'll fork and PR from that.

b-rodrigues:
> That might be my inexperience with forks, I usually push a branch and make a PR from that. I'll fork and PR from that.

I see, what you're describing is if you own the repository, but to contribute to someone else's you first must fork it.

However before moving on with this, shouldn't setting the github PAT using Sys.setenv("your pat"), before running rix::rix work?

b-rodrigues:
Another solution is to build the package locally and install it locally https://docs.ropensci.org/rix/articles/d1-installing-r-packages-in-a-nix-environment.html#installing-local-archives

I won't have much time to review PRs in the coming weeks

b-rodrigues:
OK I see, well I'll do my best to review asap, that said, until then, you can try to install the local archive instead

mihem:
So what I mean,  the lines that you want to add, these are the lines I added some time ago

https://github.com/ropensci/rix/blob/5884a9cde10fc73432e14d2801bb123c957edf37/R/fetchers.R#L515-L520

So don't think the PR is necessary. 
But haven't really looked into this deeply.

JonathanMatthewSamson:
I agree that it may be (but not my decision) out of the scope of rix to handle this; for my uses, it needs to be reproducible internally only, and access to private repositories is a must.  While that may be the case of other people who also wish to utilize rix, I'd understand if such support was not provided.

I'll put this on hold for myself until I hear whether rix wants to support this; I believe there is a way internally to use rix as-is (after the PR in its current state) with some minor code-based modification of the relevant lines.

b-rodrigues:
fixed with https://github.com/ropensci/rix/pull/585

PR #585, opened 2026-02-03
Support for private git repositories with SSH authentication

brancengregory:
Closes #599

This PR adds support for private Git repositories by introducing a `private` parameter for git packages. When `private=TRUE`, the package uses `builtins.fetchGit` with SSH authentication instead of `pkgs.fetchgit`.

This implementation addresses the limitation identified in the discussion on #559 and takes a different approach from the draft PR #560.

**Why not the GITHUB_PAT + HTTPS approach from PR #560?**

PR #560 (in draft status) attempted to use GITHUB_PAT with HTTPS URLs to access private repos. However, as noted in that PR and the issue discussion, `pkgs.fetchgit` in the Nix build sandbox cannot access SSH keys or user credentials. The generated `default.nix` files would fail during `nix-build`.

**Alternative approach: Explicit SSH URLs with `private` flag**

Instead of implicit authentication, we require users to:
1. Set `private = TRUE` in the git package list
2. Provide an SSH URL (`git@github.com:org/repo.git`)
This uses `builtins.fetchGit` which runs in user context (outside the Nix sandbox) and can access SSH keys directly. This is more reliable and explicit.

 Changes
 
- **R/nix_hash.R**: Add GITHUB_PAT authentication for downloading private repo archives during hash calculation
- **R/fetchers.R**: 
  - Add `private` parameter to `fetchgit()` and `generate_git_nix_expression()`
  - Validate SSH URL requirement when `private=TRUE`
  - Use `builtins.fetchGit` for private repos (no sha256 required)
  - Convert SSH→HTTPS for dependency resolution while keeping SSH URL in Nix expression
  - Add warning about tradeoffs when using private repos
- **R/fetchers.R** (Python): Same changes for `fetch_py_git()`

 Usage Example
 
 ```
rix(
  r_ver = "4.4.1",
  git_pkgs = list(
    package_name = "myPrivatePkg",
    repo_url = "git@github.com:myorg/private-repo.git",  # SSH URL required
    commit = "abc123",
    private = TRUE  # NEW: enables builtins.fetchGit
  )
)
```

Tradeoffs (documented in warning)

- ✅ Works seamlessly with SSH keys for private repositories
- ❌ Cannot be cached in Nix binary caches
- ❌ Requires SSH keys during evaluation
- ❌ May not work in pure Nix mode or some CI/CD environments

Notes

- This takes a more conservative/explicit approach than PR #560
- Maintains backward compatibility - all existing code works unchanged
- Only affects packages explicitly marked with private=TRUE
- Thanks to @JonathanMatthewSamson for exploring this space in PR #560 - that work helped identify the HTTPS authentication limitations

### DIFF
--- a/R/fetchers.R
+++ b/R/fetchers.R
@@ -68,7 +68,9 @@ fetchgit <- function(git_pkg, ...) {
 #' @param imports A list of packages, can be empty list
 #' @param remotes A list of remotes dependencies, can be empty list
 #' @param private Logical, if TRUE use builtins.fetchGit for private repos (uses SSH).
-#'   When TRUE, converts HTTPS URLs to SSH and skips sha256 hash requirement.
+#'   When TRUE, requires SSH repo URLs and uses them with builtins.fetchGit; SSH URLs
+#'   are temporarily converted to HTTPS only for hash calculation, while the Nix
+#'   expression continues to use the original SSH URL and skips the sha256 requirement.
 #' @return A character. Part of the Nix definition to download and build the R package
 #' from the CRAN archives.
 #' @noRd


b-rodrigues:
@JonathanMatthewSamson what do you think of this approach? This seems to solve the issue

b-rodrigues (approved):
Ok, this seems to work quite well and is a valuable addition to rix! Many thanks!

Issue #613, opened 2026-05-14
Server for buildling R packages weekly


mihem:
@b-rodrigues Sorry follow-up question to https://github.com/ropensci/rix/issues/605: since you said that building all R packakages only takes a day or so (because of all the cached packages) and storage costs would not be too high:
Don't you think it makes sense to set up a server for this?
Maybe just and old laptop that you have or since you are in public server, maybe you have the option to get a server for free?
If this is not the case, I can also ask the IT department of my university, if I may use my server for that purpose.
So then we could just use nix-build once a week (as you do right now but not for some but for all packages) and then host it via e.g. Attic (instead of cachix, which you use right now)?
Then, if everything works, all packages would build and would be available as binaries for your weekly build and the dates that you provide.

b-rodrigues:
The issue is not really setting something up technically, but setting something up that is reliable enough to be used worldwide by many users.

I cannot just self-host something like this on an old laptop or a small private server and expect it to provide the level of reliability people would expect from a public binary cache.

For our current purposes, Cachix actually works very well, so this is not really a problem at the moment.

Ideally, what we would want is for these binaries to end up in the official public NixOS cache. But that requires coordination with the NixOS infrastructure team and integration into their existing infrastructure and workflows, so it is not something that is straightforward to set up independently.

So while self-hosting with something like Attic is technically possible, the real challenge is operating and maintaining a service at the level of reliability and availability that users would expect. For now, Cachix is a good solution for our needs.

mihem:
Yeah of course, official NixOS cache would be the best solution. Did you contact them regarding this?

And yes, cachix is great. Super easy to use, and super fast and reliable, agree. But limited in space. For the dependencies of this one package I work on, cerebroAppLite, it requires 1GB (from 5GB free), although I use publc nix cache and rstats -on nix (with the bleeding edge version, it took 500MB with the latest-upstream version). You write `So only the most popular packages get built and cached,`, so how many packages do you have on rstats-on-nix cachix?

And regarding reliability: I think severs from my university are quite reliable. Not sure they could handle thousands of users .. but honestly, right now I would think there are less than 100 users? It would not replace official nix cache for R packages, only add those package not on offiical nix cache and not on rstats-on-nix I would think?

b-rodrigues:
btw, there's a new bump going on: https://raw.githack.com/b-rodrigues/nixpkgs-r-updates-fails/targets-runs/output/r-updates-fails.html

b-rodrigues:
[...]

PR #628, opened 2026-07-02
add docs about attic cache

mihem:
@b-rodrigues as you suggested here https://github.com/ropensci/rix/issues/613


b-rodrigues:
thanks!

Issue #635, opened 2026-08-22
improve docs: installing python package not on Nix


mihem:
@b-rodrigues we wrote this section about how to install python Packages no on Nix.

https://docs.ropensci.org/rix/articles/packaging-your-code.html?q=uv#part-2-python-packages-with-uv

Why do you write then at a different place that these packages should be added by changing default.nix and not using uv? https://docs.ropensci.org/rix/articles/packaging-your-code.html?q=uv#part-2-python-packages-with-uv
That's also what you wrote in your new book. In general, I though we should never change default.nix directly becaue it get's overwritten. Can rix generate this expression? Is this cleaner then using uv then?

Also I just realized that it's better to use `uv sync`   based on `pyproject.toml` because then we don't need `requirements.txt`.
I just changed this in my project here: https://github.com/mihem/checkpoint_pnp

b-rodrigues:
I don’t understand what you you mean, can you link to the exact lines?

mihem:
In rix docs and your book.


> Using Your Package in a Nix Environment
> To use your Python package in another Nix environment (e.g., for an analysis project), use pkgs.python3Packages.buildPythonPackage in your default.nix. You can point to a GitHub repository:
> ```
> pyclean = pkgs.python313Packages.buildPythonPackage rec {
>   pname = "pyclean";
>   version = "0.1.0";
>   src = pkgs.fetchgit {
>     url = "https://github.com/yourusername/pyclean";
>     rev = "commit_hash";
>     sha256 = "sha256-hash";
>   };
>   pyproject = true;
>   propagatedBuildInputs = [
>     pkgs.python313Packages.pandas
>   ];
> };
> ```
> 

But it's not a good idea to change default.nix manually.
Why not use uv here? or create a rix function that works with git, then we might not even need uv that much?

b-rodrigues:
added a new sentence that provides some guidance, thanks!

mihem:
but seriously why did we never consider adding a function in rix to add installing python pkgs via github, as we do for R?
uv is a good-enough workflow for me right now, but you often highlight that this is impure and not optimal

b-rodrigues:
>but seriously why did we never consider adding a function in rix to add installing python pkgs via github, as we do for R?

I don’t wanna do it, because I don’t need it

PR #636, opened 2026-08-22
improve python uv docs

mihem:
@b-rodrigues see  https://github.com/ropensci/rix/issues/635

`uv sync` is better because we can remove `requirements.txt` and have only one source of truth.
`uv init` without specify python version here is better because we also specified the version in `py_conf` and again we only have one source of truth, not two, which could drift apart.

### DIFF
--- a/vignettes/installing-r-packages.Rmd
+++ b/vignettes/installing-r-packages.Rmd
@@ -428,20 +428,25 @@ to install packages. This is also useful if you work on a project with colleague
 use `uv` and that don’t want (yet) to use Nix. `uv`, is 10-100x faster than `pip` and
 also generates a lock file for improved reproducibility.
 
-The idea is to install `uv` in your shell (but not any Python nor Python packages):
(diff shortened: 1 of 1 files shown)


vignettes/installing-r-packages.Rmd
```diff
@@ -428,39 +428,44 @@ to install packages. This is also useful if you work on a project with colleague
...
 
-The idea is to install `uv` in your shell (but not any Python nor Python packages):
+The idea is to install `uv` alongside the Python version specified in `py_conf`,
+but let `uv` install the project dependencies:
 
 ```{r, eval = F}
 rix(
   ...
+  py_conf = list(
+    py_version = "3.13",
```
b-rodrigues, line 438: shouldn't python come from uv? why a nix-managed python?

mihem: Yeah you are probably right if we use uv for the other pkgs then we can also use it for python itself. Do you change this or do you want me to change this?

b-rodrigues:
thanks!

